PR #1, opened 2020-06-17
Initial Project

TrustHenry:
This is the initial configuration of the stoa API project,
which operates in a Node.js environment and is implemented with TypeScript.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,4 +27,3 @@ jobs:
     - run: npm ci
     - run: npm run build --if-present
     - run: npm test
-      
--- a/README.md
+++ b/README.md
@@ -1,2 +1,28 @@
 # stoa
 API server for Agora
+
+
+
+## Build instructions
+
+# Install the node.Js >= 12.18.0 (includes npm 6.14.4) (you might want to use a newer version)
+
+
+# Building on Ubuntu
+# Ubuntu is already nodejs installed.
+$ sudo apt-get install nodejs
+$ sudo apt-get install npm
+$ npm ci
+
+# Building on MacOS
+$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+$ brew install node
+$ npm ci
+
+## How to start
+- `npm start`: Run the stoa in the Node.Js. It is being watched by nodemon.
+               When changing the source, the code configured with typescript
+               is transpiled to javascript and restarted.
+
+## Stoa test
+- `npm test`: Test run the test.ts files in the tests folder.
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -1,0 +1,20 @@
+import express from "express";
+
+class Stoa {
+    public stoa: express.Application;
+
+    constructor () {
+        this.stoa = express();
+
+        this.stoa.get("/validators",
+            (req: express.Request, res: express.Response, next: express.NextFunction) => {
+            res.send(req.query.height);
+        });
+
+        this.stoa.get("/validator/:address",
+            (req: express.Request, res: express.Response, next: express.NextFunction) => {
+            res.send(req.params.address + ':' + req.query.height);
+        });
+    }
+}
+export default Stoa;
--- a/tests/test.test.ts
+++ b/tests/test.test.ts
@@ -1,0 +1,4 @@
+describe('Calculator', () => {
+    it('nothing', () => {
+    });
+  });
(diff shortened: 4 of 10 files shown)


Geod24:
Please also add a CI ;)

Geod24:
https://github.com/actions/starter-workflows/blob/master/ci/node.js.yml

MichaelKim20 (approved):

src/Stoa.ts
```diff
@@ -0,0 +1,20 @@
...
+
+class Stoa {
+    public stoa: express.Application;
+
+    constructor () {
+        this.stoa = express();
+
+        this.stoa.get("/validators",
+            (req: express.Request, res: express.Response, next: express.NextFunction) => {
+            res.send(req.query.height);
```
AndrejMitrovic, line 11: Where does the `height` property come from? Is the support code for this automatically generated?

Also, height should be an optional parameter. Does this code handle that well?

tests/test.test.ts
```diff
@@ -0,0 +1,4 @@
+describe('Calculator', () => {
```
AndrejMitrovic, line 1: Accidentally committed this?

tsconfig.json
```diff
@@ -0,0 +1,72 @@
...
+    // "strictBindCallApply": true,           /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
+    // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+    // "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
+    // "alwaysStrict": true,                  /* Parse in strict mode and emit "use strict" for each source file. */
+
+    /* Additional Checks */
+    // "noUnusedLocals": true,                /* Report errors on unused locals. */
+    // "noUnusedParameters": true,            /* Report errors on unused parameters. */
+    // "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
+    // "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
```
AndrejMitrovic: Is there a reason why these two are not enabled?

src/Stoa.ts
TrustHenry: In the get method request, the query parameter is stored in a dynamic object and this is an option.
IF processing will be done in logic construction in the future.

tests/test.test.ts
TrustHenry: Here's a sample of test implementation for CI.

src/Stoa.ts
AndrejMitrovic: Alright!

tsconfig.json
TrustHenry: Good catch! It's a good option for error detection.

tests/test.test.ts
AndrejMitrovic: Aha, I see it in the CI. Alright.

.github/workflows/main.yml
```diff
@@ -27,4 +27,3 @@ jobs:
     - run: npm ci
     - run: npm run build --if-present
     - run: npm test
-
```
AndrejMitrovic, line 30: This isn't really that important, but I suggest that you and Michael have the same vs.code settings when it comes to whitespace and newlines. Otherwise whenever someone makes a PR there will always be these unrelated changes.

AndrejMitrovic (approved):

README.md
```diff
@@ -1,2 +1,28 @@
 # stoa
 API server for Agora
+
+
+
+## Build instructions
+
+# Install the node.Js >= 12.18.0 (includes npm 6.14.4) (you might want to use a newer version)
```
AndrejMitrovic, line 8: @TrustHenry the readme says it requires node.js >= 12.18.0.

So then it's unnecessary to test node.js v10, right? We might want to disable it from the CI.

.github/workflows/main.yml
TrustHenry: That's right. 
This is a blank mistake when I set up my initial CI.

README.md
TrustHenry: We actually need to test the previous version.
We recommend more than the latest version here, but I would like to describe it as a guideline.
When I tried, Node 8 found a problem with the mocha test.
I want to support at least 10 version.

AndrejMitrovic: Oh ok, that's good!

Geod24 (approved):

PR #4, opened 2020-06-18
Add body-parser node_module and add post /push URL

TrustHenry:


package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "@types/sqlite3": "^3.1.6",
     "assert": "^2.0.0",
+    "body-parser": "^1.19.0",
```
MichaelKim20, line 25: It would be better to separate the module into a different commit.

TrustHenry: I've separated it with two commits

MichaelKim20 (approved):
I got the following satisfactory test result.
https://gist.github.com/MichaelKim20/02678cf1eec3f756e2b1af09440db096

Looks good to me

src/Stoa.ts
```diff
@@ -1,10 +1,15 @@
 import express from "express";
+import bodyParser from "body-parser";
 
 class Stoa {
     public stoa: express.Application;
 
     constructor () {
         this.stoa = express();
+        // parse application/x-www-form-urlencoded
+        this.stoa.use(bodyParser.urlencoded({ extended: false }))
```
Geod24, line 10: That's for the `height` parameter right ?

TrustHenry: I don't think so. 
That is Returns middleware that only parses  and Content-Type header matches and accepts only UTF-8 encoding

PR #5, opened 2020-06-18
Implement the ability to input and output blocks into a database

MichaelKim20:
Implement the class that inserts and reads the block into the database.

### DIFF
--- a/src/modules/storage/BlockStorage.ts
+++ b/src/modules/storage/BlockStorage.ts
@@ -1,0 +1,152 @@
+/*******************************************************************************
+
+    The class that insert and read the block into the database.
+
+    Now only block header are stored.
+    Transactions and other data will be managed by different classes.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import * as sqlite from 'sqlite3';
+
+/**
+ * The class that insert and read the block into the database.
+ */
+class BlockStorage
+{
+    /**
+     *  The instance of sqlite3
+     */
+    private db: sqlite.Database;
+
+    /**
+     * Constructor
+     * @param filename Valid values are filenames,
+     * ":memory:" for an anonymous in-memory database and
+     * an empty string for an anonymous disk-based database
+     * @param callback : If provided, this function will be called when
+     * the database was opened successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    constructor (filename : string, callback : any = null)
+    {
+        this.db = new sqlite.Database(filename,
+            sqlite.OPEN_CREATE | sqlite.OPEN_READWRITE |
+            sqlite.OPEN_SHAREDCACHE);
+        this.db.configure("busyTimeout", 1000);
+        this.createTable(callback);
+    }
+
+    /**
+     * Creates a table. table name is `block`.
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    public createTable (callback : any = null)
+    {
+        var sql =
+        `CREATE TABLE IF NOT EXISTS blocks
+        (
+            height INTEGER PRIMARY KEY,
+            prev_block TEXT,
+            validators TEXT,
+            merkle_root TEXT,
+            signature TEXT
+        )`;
+        this.db.run(sql, (err : any) =>
+        {
+            if (callback != null)
+                callback(err);
+        });
+    }
+
+    /**
+     * Puts a block to database
+     * @param data a block data
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    public put (data : any, callback : any = null)
+    {
+        if (
+                (data == null) ||
+                (data.header == undefined) ||
+                (data.header.height == undefined) ||
+                (data.header.height.value == undefined) ||
+                (data.header.prev_block == undefined) ||
+                (data.header.validators == undefined) ||
+                (data.header.validators._storage == undefined) ||
+                (data.header.merkle_root == undefined) ||
+                (data.header.signature == undefined)
+        ) {
+            if (callback != null)
+                callback("Parameter validation failed.");
+            return;
+        }
+
+        var sql =
+        `INSERT INTO blocks
+            (height, prev_block, validators, merkle_root, signature)
+        VALUES
+            (?, ?, ?, ?, ?)
+        ON CONFLICT(height) DO UPDATE SET
+            prev_block = prev_block,
+            validators = validators,
+            merkle_root = merkle_root,
+            signature = signature`;
+        this.db.run(sql,
+            [
+                data.header.height.value,
+                data.header.prev_block,
+                JSON.stringify(data.header.validators._storage),
+                data.header.merkle_root,
+                data.header.signature
+            ], (err : any) =>
+        {
+            if (callback != null)
+                callback(err);
+        });
+    }
+
+    /**
+     * Gets a block data
+     * @param height the height of the block to get
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     * The second argument is result set.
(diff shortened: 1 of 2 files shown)


src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,128 @@
...
+        this.createTable(callback);
+    }
+
+    /**
+     * Creates a table. table name is `block`.
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    public createTable (callback : any = null)
```
TrustHenry, line 53:
```suggestion
    public createTables (callback : any = null)
```
Or setTables Or initialTables

I think we will define the table schema here.
transaction tables as well as block tables ..

MichaelKim20: When the constructor runs, it creates a table if it does not exist. Therefore, when an API server is installed, there is no need to do additional work on table creation.

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,128 @@
...
+    {
+        var sql =
+        `CREATE TABLE IF NOT EXISTS block
+        (
+            height INTEGER PRIMARY KEY,
+            prev_block TEXT,
+            merkle_root TEXT,
+            signature TEXT
```
TrustHenry, line 62: It would be better to add the `validators` field.

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,128 @@
...
+     * The first argument is an error object.
+     */
+    public put (data : any, callback : any = null)
+    {
+        var sql =
+        `INSERT INTO block
+            (height, prev_block, merkle_root, signature)
+        VALUES
+            (?, ?, ?, ?)
+        ON CONFLICT(height) DO UPDATE SET
```
TrustHenry, line 101: Is there a scenario where it is possible to update the data if it already exists?

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Creates a table. table name is `block`.
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    public createTable (callback : any = null)
+    {
+        var sql =
+        `CREATE TABLE IF NOT EXISTS block
```
TrustHenry:
```suggestion
        `CREATE TABLE IF NOT EXISTS blocks
```

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,128 @@
...
+     * Gets a block data
+     * @param height the height of the block to get
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     * The second argument is result set.
+     */
+    public get (height : number, callback : any = null)
+    {
+        this.db.all("SELECT * FROM block WHERE height = ?",
```
TrustHenry: We'd better define the column and bring it in.
Not `SELECT *`
Unnecessary information may be added when returning to API after inquiry.

src/modules/storage/BlockStorage.ts
MichaelKim20: Okay. I'll look into it.

MichaelKim20: I added it

src/modules/storage/BlockStorage.ts
MichaelKim20: This shouldn't have happened.
However, if this happens, it is to avoid the error caused by it.
If this can happen, data recovery by polling may be added in the future.

https://www.sqlite.org/lang_UPSERT.html

src/modules/storage/BlockStorage.ts
MichaelKim20: It would be better if we expand it in the future.

src/modules/storage/BlockStorage.ts
MichaelKim20: Good point

src/modules/storage/BlockStorage.ts
MichaelKim20: Good point!

src/modules/storage/BlockStorage.ts
MichaelKim20: I changed it.

src/modules/storage/BlockStorage.ts
MichaelKim20: I changed it.

AndrejMitrovic:
Was this something @geod24 requested?

MichaelKim20:
> Was this something @Geod24 requested?

I think this is one of the necessary functions. So I made it.

AndrejMitrovic:
Right, I was looking at it through my phone and just assumed from the title that the entire blocks were stored. I assumed wrong.

src/modules/storage/BlockStorage.ts
Geod24: Then the method shouldn't be `public`. It shouldn't be used inside the class. And it is not generic, so the name should reflect that. E.g. `createBlocksTable`

TrustHenry (approved):
LGTM

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Gets a block data
+     * @param height the height of the block to get
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     * The second argument is result set.
+     */
+    public get (height : number, callback : any = null)
```
Geod24, line 128: The callback seems to have a defined type, doesn't it ?

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,152 @@
...
+            [
+                data.header.height.value,
+                data.header.prev_block,
+                JSON.stringify(data.header.validators._storage),
+                data.header.merkle_root,
+                data.header.signature
+            ], (err : any) =>
+        {
+            if (callback != null)
+                callback(err);
```
Geod24, line 116: Same here, it has a defined type

src/modules/storage/BlockStorage.ts
```diff
@@ -0,0 +1,152 @@
...
+    public createTable (callback : any = null)
+    {
+        var sql =
+        `CREATE TABLE IF NOT EXISTS blocks
+        (
+            height INTEGER PRIMARY KEY,
+            prev_block TEXT,
+            validators TEXT,
+            merkle_root TEXT,
+            signature TEXT
```
Geod24, line 62: Some constraints missing (e.g. `NOT NULL`). Also what schema do you have in mind ?

tests/BlockStorage.test.ts
```diff
@@ -0,0 +1,358 @@
...
+ *  Start test
+ */
+function runBlockStorageTest ()
+{
+    create();
+}
+
+describe('BlockStorage', () =>
+{
+    it('BlockStorage', () =>
```
Geod24, line 354: Please give it an actual description. This is supposed to read as a sentence.

Geod24 (changes requested):
Few things

TrustHenry:
We applied the review contents and are currently developing a common module with Storage.

src/modules/storage/BlockStorage.ts
MichaelKim20: I was planning to revise it when it was more specific.

src/modules/storage/BlockStorage.ts
MichaelKim20: If there is a value to be returned, the second parameter is valid.

tests/BlockStorage.test.ts
MichaelKim20: Yeah, I think we should be more specific. ex) `Tests on the creation, writing, and reading of BlockStorage.`

src/modules/storage/BlockStorage.ts
Geod24: That doesn't relate to the question

PR #10, opened 2020-06-22
Implement a class that makes it easier to calculate the hash

MichaelKim20:
I Implemented a class that makes it easier to calculate the hash.
This was implemented to a UTXO key using the transaction's hash and
output index.

### DIFF
--- a/tests/hash.test.ts
+++ b/tests/hash.test.ts
@@ -1,0 +1,76 @@
+/*******************************************************************************
+
+    Test that create hash.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import * as assert from 'assert';
+import * as hm from '../src/modules/common/Hash'
+const hms = hm.default;
+
+describe('Hash', () => {
+    // Buffer has the same content. However, when printed with hex strings,
+    // the order of output is different.
+    // This was treated to be the same as D language.
+    it('Test of reading and writing hex string', () => {
+        // Read from hex string
+        let h =
+            new hms.Hash('0x5d7f6a7a30f7ff591c8649f61eb8a35d034824ed5cd252c2c6f10cdbd223671' +
+                    '3dc369ef2a44b62ba113814a9d819a276ff61582874c9aee9c98efa2aa1f10d73');
+
+        // Check
+        assert.equal(h.toHexString(),
+            '0x5d7f6a7a30f7ff591c8649f61eb8a35d034824ed5cd252c2c6f10cdbd2236713d' +
+            'c369ef2a44b62ba113814a9d819a276ff61582874c9aee9c98efa2aa1f10d73');
+    });
+
+    it('Test of hash', () => {
+        // Create Hash class
+        let h = new hms.Hash();
+
+        // Hash
+        h.hash(Buffer.from("abc"));
+
+        // Check
+        assert.equal(h.toHexString(),
+            '0x239900d4ed8623b95a92f1dba88ad31895cc3345ded552c22d79ab2a39c5877' +
+            'dd1a2ffdb6fbb124bb7c45a68142f214ce9f6129fb697276a0d4d1c983fa580ba');
+    });
+
+    // https://github.com/bpfkorea/agora/blob/2d758a693f9df376f9b873f7c7897c0787b582f1/source/agora/common/Hash.d#L260-L265
+    it('Test of multi hash', () => {
+        // Source 1 : "foo"
+        let foo = new hms.Hash();
+        foo.hash(Buffer.from("foo"));
+
+        // Source 2 : "bar"
+        let bar = new hms.Hash();
+        bar.hash(Buffer.from("bar"));
+
+        // Hash Multi
+        let h = new hms.Hash();
+        h.hashMulti(foo.buffer, bar.buffer);
+
+        // Check
+        assert.equal(h.toHexString(),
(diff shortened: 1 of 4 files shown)


tests/hash.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    it('Test of utxo key, using makeUTXOKey', () => {
+        let hash = new hms.Hash();
+        hash.makeUTXOKey('0x5d7f6a7a30f7ff591c8649f61eb8a35d034824ed5cd252c2c6f10cdbd223671' +
+        '3dc369ef2a44b62ba113814a9d819a276ff61582874c9aee9c98efa2aa1f10d73', 1);
+
+        assert.equal(hash.toHexString(),
+            '0x73f6edb2e400bb6ed3b109a110bac990cbfd845844f3bcc4d716d49653be0490' +
+            '4982d19416b0752fe2eb18095ebd5a984b4acf1cfe951b5ee1aef98c2617ab7e');
+    });
```
Geod24: Trying this in Agora I got `0x7c95c29b184e47fbd32e58e5abd42c6e22e8bd5a7e934ab049d21df545e09c2e33bb2b89df2e59ee01eb2519b1508284b577f66a76d42546b65a6813e592bb84`

```diff
diff --git a/source/agora/consensus/data/UTXOSetValue.d b/source/agora/consensus/data/UTXOSetValue.d
index e134369e..625c0408 100644
--- a/source/agora/consensus/data/UTXOSetValue.d
+++ b/source/agora/consensus/data/UTXOSetValue.d
@@ -149,5 +149,8 @@ version (unittest) public class TestUTXOSet

 unittest
 {
+    import std.stdio;
+    writeln(UTXOSetValue.getHash(Hash("0x5d7f6a7a30f7ff591c8649f61eb8a35d034824ed5cd252c2c6f10cdbd223671" ~
+                                      "3dc369ef2a44b62ba113814a9d819a276ff61582874c9aee9c98efa2aa1f10d73"), 1));
     testSymmetry!UTXOSetValue();
 }
```

tests/hash.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        // Hash Multi
+        let hash = new hms.Hash();
+        hash.hashMulti(tx_hash_buffer, idx_buffer);
+
+        // Check
+        assert.equal(hash.toHexString(),
+            '0x73f6edb2e400bb6ed3b109a110bac990cbfd845844f3bcc4d716d49653be0490' +
+            '4982d19416b0752fe2eb18095ebd5a984b4acf1cfe951b5ee1aef98c2617ab7e');
+    });
```
Geod24: There's no point in having this test, just test `makeUTXOKey`.

tests/hash.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        // Hash Multi
+        let h = new hms.Hash();
+        h.hashMulti(foo.buffer, bar.buffer);
+
+        // Check
+        assert.equal(h.toHexString(),
+            '0xe0343d063b14c52630563ec81b0f91a84ddb05f2cf05a2e4330ddc79bd3a06e57' +
+            'c2e756f276c112342ff1d6f1e74d05bdb9bf880abd74a2e512654e12d171a74');
+    });
```
Geod24, line 65: This one is good though 👍

tests/hash.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        // Create Hash class
+        let h = new hms.Hash();
+
+        // Hash
+        h.hash(Buffer.from("foo"));
+
+        // Check
+        assert.equal(h.toHexString(),
+            '0x8d8fd1dd408231aba0d7bbc59d2d98bb14c724843720e13707aeb43dfb23e7c16' +
+            'd5d1854b343ffc2d6cab73b7ad9519dd73f53a6566aa4846b3e9de6302300ca');
```
Geod24:
```suggestion
        // Test with https://tools.ietf.org/html/rfc7693#appendix-A
        h.hash(Buffer.from("abc"));
        assert.equal(h.toHexString(),
            '0x239900d4ed8623b95a92f1dba88ad31895cc3345ded552c22d79ab2a39c5877'
            'dd1a2ffdb6fbb124bb7c45a68142f214ce9f6129fb697276a0d4d1c983fa580ba');
```

It's better to use the RFC for this.

tests/hash.test.ts
MichaelKim20: There's a data type called `bigint` in the typescript.

Geod24: I don't understand what you mean.

tests/hash.test.ts
MichaelKim20: I fixed

tests/hash.test.ts
MichaelKim20: I fixed

tests/hash.test.ts
MichaelKim20: I changed

tests/hash.test.ts
MichaelKim20: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

MichaelKim20: When we use the data type defined as 'ulong' in D language, it will be helpful to refer to the link above.

Geod24: What I mean is:
Remove the test. Because the test is redundant. The code in the test is a copy of the code in `makeUTXOKey`.
The next test tests `makeUTXOKey`. We only need the next test.

MichaelKim20: Okay

TrustHenry (approved):
LGTM

TrustHenry:
I think it would be good to be static in the future.

Geod24 (approved):

PR #11, opened 2020-06-23
Refactoring modules (BlockStorage, Hash)

MichaelKim20:
I changed the export form of modules.
I changed the code style.

### DIFF
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -329,8 +329,9 @@ function putAllBlockData ()
 /**
  * Gets one block data
  */
-function getBlockData (height : number)
+function getBlockData (height: any)
 {
+    block_storage.get(height);
     var res = block_storage.get(height, (err:any, rows:any) =>
     {
         assert.equal(rows.length, 1);
@@ -351,7 +352,7 @@ function runBlockStorageTest ()
 
 describe('BlockStorage', () =>
 {
-    it('BlockStorage', () =>
+    it('Test block storage and inquiry function.', () =>
     {
         runBlockStorageTest();
     });
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -12,7 +12,7 @@
 *******************************************************************************/
 
 import * as assert from 'assert';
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Refactoring means changing the code without changing the functionality. But the PR introduces a database schema change.

tests/Storage.test.ts
```diff
@@ -309,18 +309,19 @@ function putAllBlockData ()
     var idx = 0;
     var doPut = () =>
     {
+        if (idx >= sample_data.length)
+        {
+            getBlockData(1);
+                return;
+        }
+
```
Geod24: This comment is on the first commit. You should add a commit description.
The commit description needs to explain why this change is necessary.
The diff only shows some code being moved. If there was a problem, it should be described.

MichaelKim20: I will add this to the next PR. Thanks.

MichaelKim20:
> Refactoring means changing the code without changing the functionality. But the PR introduces a database schema change.

I will separate the commit into three PRs.
So I will solve this problem.

tests/Storage.test.ts
```diff
@@ -329,8 +329,9 @@ function putAllBlockData ()
 /**
  * Gets one block data
  */
-function getBlockData (height : number)
+function getBlockData (height: any)
```
Geod24, line 332: Why `any` ?

MichaelKim20: I want to process this as a string.
The data type of 'height' is  `ulong` and it is difficult to process in TypeScript. 
Therefore, it is for processing as a string.
It should use `bigint` to be exact, but it would be more efficient not to convert what can be treated as a string.

tests/Storage.test.ts
MichaelKim20: https://github.com/bpfkorea/stoa/pull/12
I created a new PR for this.

tests/Storage.test.ts
Geod24: I see. I am not a fan of this approach, because numbers should be stored as numbers, not string, but if it's easier in TypeScript, let's go with this.

PR #12, opened 2020-06-23
Fix the test code (putAllBlockData)

MichaelKim20:
I corrected the problem in function `putAllBlockData` when the size of `sample_data` is 0.

### DIFF
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -309,17 +309,18 @@ function putAllBlockData ()
     var idx = 0;
     var doPut = () =>
     {
+        if (idx >= sample_data.length)
+        {
+            getBlockData(1);
+            return;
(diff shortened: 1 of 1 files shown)


tests/Storage.test.ts
```diff
@@ -309,18 +309,19 @@ function putAllBlockData ()
     var idx = 0;
     var doPut = () =>
     {
+        if (idx >= sample_data.length)
+        {
+            getBlockData(1);
+                return;
```
Geod24:
```suggestion
            getBlockData(1);
             return;
```

Also please add the explanation to the commit **message**, not just the PR description.

MichaelKim20: Oh, I need to fix it, thank you.

PR #13, opened 2020-06-23
Create and apply a storages superclass

TrustHenry:
Manage DB Instances and reduce code duplication.

This is the pre-requisite for further PR.

### DIFF
--- a/src/modules/storage/BlockStorage.ts
+++ b/src/modules/storage/BlockStorage.ts
@@ -14,34 +14,22 @@
 
 *******************************************************************************/
 
-import * as sqlite from 'sqlite3';
+import { Storages } from './Storages';
 
 /**
  * The class that insert and read the block into the database.
  */
-export class BlockStorage
+export class BlockStorage extends Storages
 {
     /**
-     *  The instance of sqlite3
-     */
-    private db: sqlite.Database;
-
-    /**
      * Constructor
-     * @param filename Valid values are filenames,
-     * ":memory:" for an anonymous in-memory database and
-     * an empty string for an anonymous disk-based database
      * @param callback : If provided, this function will be called when
      * the database was opened successfully or when an error occurred.
      * The first argument is an error object.
      */
-    constructor (filename: string, callback?: any)
+    constructor (callback?: any)
     {
-        this.db = new sqlite.Database(filename,
-            sqlite.OPEN_CREATE | sqlite.OPEN_READWRITE |
-            sqlite.OPEN_SHAREDCACHE);
-        this.db.configure("busyTimeout", 1000);
-        this.createTable(callback);
+        super(callback);
     }
 
     /**
@@ -142,11 +130,4 @@ export class BlockStorage
         });
     }
 
-    /**
-     * Closes database
-     */
-    public close ()
-    {
-        this.db.close();
-    }
 }
--- a/src/modules/storage/Storages.ts
+++ b/src/modules/storage/Storages.ts
@@ -1,0 +1,54 @@
+/*******************************************************************************
+
+    The superclass of storages.
+
+    This has a DB instance.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import * as sqlite from 'sqlite3';
+
+export class Storages
+{
+    /**
+     *  The instance of sqlite3
+     */
+    protected db: sqlite.Database;
+
+    /**
+     * Constructor
+     * @param filename Valid values are filenames,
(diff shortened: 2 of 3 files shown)


src/modules/storage/BlockStorage.ts
```diff
@@ -14,34 +14,23 @@
...
      */
-    constructor (filename: string, callback?: any)
+    constructor (callback: any = null)
     {
-        this.db = new sqlite.Database(filename,
-            sqlite.OPEN_CREATE | sqlite.OPEN_READWRITE |
-            sqlite.OPEN_SHAREDCACHE);
-        this.db.configure("busyTimeout", 1000);
-        this.createTable(callback);
+        super(undefined, callback);
```
AndrejMitrovic: Am I missing something here? You are calling the base constructor with `undefined`?

src/modules/storage/Storages.ts
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Constructor
+     * @param filename Valid values are filenames,
+     * ":memory:" for an anonymous in-memory database and
+     * an empty string for an anonymous disk-based database
+     * @param callback : If provided, this function will be called when
+     * the database was opened successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    constructor (filename : string = ".cache\ledger", callback : any = null)
```
AndrejMitrovic: filename should not be a default parameter if you want this class to be reusable. What if you forget to set the filename? then you end up overwriting the ledger storage by mistake.

src/modules/storage/BlockStorage.ts
TrustHenry: It seems unnecessary to me, too.
DB filename can be fixed.

src/modules/storage/Storages.ts
TrustHenry: Thank you reviews.
^ Fixed the DB filename.

src/modules/storage/Storages.ts
```diff
@@ -0,0 +1,54 @@
...
+     * @param filename Valid values are filenames,
+     * ":memory:" for an anonymous in-memory database and
+     * an empty string for an anonymous disk-based database
+     * @param callback : If provided, this function will be called when
+     * the database was opened successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    constructor (callback: any = null)
+    {
+        let filename: string = ".cache";
```
AndrejMitrovic, line 36: I still don't understand what you're trying to accomplish.

You will have multiple instances of `Storages` all writing to the same file?

TrustHenry: Yes, multiple Storages are managed in a single file.
I'd like to join a number of tables and use them.

TrustHenry: Is it because the name storage is strange?

AndrejMitrovic: Then why do you need multiple instances of `sqlite.Database`? If you will only use one file then you only need one database handle.

MichaelKim20: I think this is what @AndrejMitrovic  wants.
I think this is good too.

```suggestion
    constructor (filename: string, callback?: any)
```

AndrejMitrovic: Yes that is what I thought Henry tried to do at first. But according to him he wants a single file for all the tables..

src/modules/storage/Storages.ts
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Constructor
+     * @param filename Valid values are filenames,
+     * ":memory:" for an anonymous in-memory database and
+     * an empty string for an anonymous disk-based database
+     * @param callback : If provided, this function will be called when
+     * the database was opened successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    constructor (callback?: any)
```
MichaelKim20, line 34:
```suggestion
    constructor (filename: string, callback?: any)
```

src/modules/storage/Storages.ts
AndrejMitrovic: I'll leave design decisions to you two. It was just a comment.

TrustHenry: Maybe @AndrejMitrovic  talking about a single-ton design pattern. Is that right?

TrustHenry: Then it should get a DB Instance through `getDbInstancea`
In subsequent operations, you can manage instances as one after creating superclass in `Stoa.ts`

This PR is still applied only to the test class.

src/modules/storage/Storages.ts
TrustHenry: I don't think it's necessary yet.

MichaelKim20 (approved):
LGTM

PR #14, opened 2020-06-23
Store JSON data of API POST 'push'

TrustHenry:
Parse JSON data and save it in block storage.
This operation will be asynchronous.

Columns added in the block table due to table schema structure change.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -1,15 +1,27 @@
 import express from "express";
 import bodyParser from "body-parser";
+import { LedgerStorage } from "./modules/storage/LedgerStorage";
 
 class Stoa {
     public stoa: express.Application;
 
-    constructor () {
+    public ledger_storage: LedgerStorage;
+
+    constructor (file_name: string) {
         this.stoa = express();
         // parse application/x-www-form-urlencoded
         this.stoa.use(bodyParser.urlencoded({ extended: false }))
         // parse application/json
         this.stoa.use(bodyParser.json())
+        // create blockStorage
+        this.ledger_storage = new LedgerStorage(file_name, (err: Error | null) =>
+        {
+            if (err != null)
+            {
+                console.error(err);
+                throw new Error(err.message);
+            }
+        });
 
         this.stoa.get("/validators",
             (req: express.Request, res: express.Response, next: express.NextFunction) => {
@@ -21,10 +33,43 @@ class Stoa {
             res.send(req.params.address + ':' + req.query.height);
         });
 
+        /**
+         * When a request is received through the `/push` handler
+         * JSON block data is parsed and stored on each storage.
+         */
         this.stoa.post("/push",
             (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            console.log(req.body);
-            res.status(200).send();
+
+            var block: any;
+            if (req.body.block == undefined)
+            {
+                res.status(400).send("Missing 'block' object in body");
+                return;
+            }
+
+            try {
+                block = JSON.parse(req.body.block);
+            } catch(e) {
+                res.status(400).send("Not a valid JSON format");
+                return;
+            }
+
+            console.log(block);
+
+            this.ledger_storage.putBlocks(block, (err: any) =>
+            {
+                if (err != null)
+                {
+                    console.error("Failed to store the payload of a push to the DB: " + err);
+                    res.status(500).send("An error occurred while saving");
+                    return;
+                }
+                else
+                {
+                    res.status(200).send();
+                    return;
+                }
+            });
         });
     }
 }
--- a/src/main.ts
+++ b/src/main.ts
@@ -3,7 +3,7 @@ import Stoa from './Stoa';
 import express from "express";
 
 const port: number = Number(process.env.PORT) || 3836;
-const stoa: express.Application = new Stoa().stoa;
+const stoa: express.Application = new Stoa("database").stoa;
 
 stoa.listen(port, () => console.log(`Express server listening at ${port}`))
 .on('error', err => console.error(err));
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ dist
 
 # webpack build ouputs
 .dist
+
+# database
+database
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    The class that insert and read the block into the database.
+    The class that create and insert and read the ledger into the database.
 
     Now only block header are stored.
     Transactions and other data will be managed by different classes.
@@ -17,30 +17,32 @@
 import { Storages } from './Storages';
 
 /**
- * The class that insert and read the block into the database.
+ * The class that insert and read the ledger into the database.
  */
-export class BlockStorage extends Storages
+export class LedgerStorage extends Storages
 {
     /**
-     * Creates a table. table name is `block`.
+     * Creates tables related to the ledger.
      * @param callback If provided, this function will be called when
      * the database was finished successfully or when an error occurred.
      * The first argument is an error object.
      */
-    public createTable (callback?: any)
+    public createTables (callback: (err: Error | null) => void)
     {
         var sql =
         `CREATE TABLE IF NOT EXISTS blocks
         (
-            height INTEGER PRIMARY KEY,
-            prev_block TEXT,
-            validators TEXT,
-            merkle_root TEXT,
-            signature TEXT
+            height INTEGER NOT NULL PRIMARY KEY,
+            prev_block TEXT NOT NULL,
+            validators TEXT NOT NULL,
+            merkle_root TEXT NOT NULL,
+            signature TEXT,
+            tx_count INTEGER NOT NULL,
+            enrollment_count INTEGER NOT NULL
         )`;
-        this.db.run(sql, (err: any) =>
+        this.db.run(sql, (err: Error | null) =>
         {
-            if (callback != undefined)
+            if (callback != null)
                 callback(err);
         });
     }
@@ -52,7 +54,7 @@ export class BlockStorage extends Storages
      * the database was finished successfully or when an error occurred.
      * The first argument is an error object.
      */
-    public put (data: any, callback?: any)
+    public putBlocks (data: any, callback: (err: Error | null) => void)
     {
         if (
                 (data == null) ||
@@ -63,33 +65,34 @@ export class BlockStorage extends Storages
                 (data.header.validators == undefined) ||
                 (data.header.validators._storage == undefined) ||
                 (data.header.merkle_root == undefined) ||
-                (data.header.signature == undefined)
+                (data.header.signature == undefined) ||
+                (data.txs == undefined)
         ) {
-            if (callback != undefined)
-                callback("Parameter validation failed.");
+            if (callback != null)
+                callback(new Error("Parameter validation failed."));
             return;
         }
 
+        var tx_count: number = data.txs.length;
+        var enrollment_count: number = data.enrollments? data.enrollments.length: 0;
+
         var sql =
-        `INSERT INTO blocks
-            (height, prev_block, validators, merkle_root, signature)
-        VALUES
-            (?, ?, ?, ?, ?)
-        ON CONFLICT(height) DO UPDATE SET
-            prev_block = prev_block,
-            validators = validators,
-            merkle_root = merkle_root,
-            signature = signature`;
+            `INSERT INTO blocks
+                (height, prev_block, validators, merkle_root, signature, tx_count, enrollment_count)
+            VALUES
+                (?, ?, ?, ?, ?, ?, ?)`;
         this.db.run(sql,
             [
                 data.header.height.value,
                 data.header.prev_block,
                 JSON.stringify(data.header.validators._storage),
                 data.header.merkle_root,
-                data.header.signature
-            ], (err: any) =>
+                data.header.signature,
+                tx_count,
+                enrollment_count
+            ], (err: Error | null) =>
         {
-            if (callback != undefined)
+            if (callback != null)
                 callback(err);
         });
     }
@@ -102,20 +105,17 @@ export class BlockStorage extends Storages
      * The first argument is an error object.
      * The second argument is result set.
      */
-    public get (height: any, callback?: any)
+    public getBlocks (height: any, callback: (err: Error | null, rows: any[]) => void)
     {
         var sql =
         `SELECT
-            height, prev_block, validators, merkle_root, signature
+            height, prev_block, validators, merkle_root, signature, tx_count, enrollment_count
         FROM
             blocks
         WHERE height = ?`;
-        this.db.all(sql, [height], (err: any, rows: any) =>
+        this.db.all(sql, [height], (err: Error | null, rows: any[]) =>
         {
-            if (callback != undefined)
-                callback(err, rows);
-            else
-                console.log(callback);
+            callback(err, rows);
         });
     }
 
--- a/src/modules/storage/Storages.ts
+++ b/src/modules/storage/Storages.ts
@@ -28,33 +28,30 @@ export class Storages
      * ":memory:" for an anonymous in-memory database and
      * an empty string for an anonymous disk-based database
      */
-    constructor (filename: string, callback?: any)
+    constructor (filename: string, callback: (err: Error | null) => void)
     {
         this.db = new sqlite.Database(filename,
             sqlite.OPEN_CREATE | sqlite.OPEN_READWRITE |
-            sqlite.OPEN_SHAREDCACHE, (err: any) =>
+            sqlite.OPEN_SHAREDCACHE, (err: Error | null) =>
             {
                 if (err != null)
-                {
-                    if (callback != undefined)
-                        callback(err);
-                }
+                    callback(err);
 
                 this.db.configure("busyTimeout", 1000);
-                this.createTable((err: any) => {
-                    if (callback != undefined)
+                this.createTables((err: Error | null) => {
+                    if (callback != null)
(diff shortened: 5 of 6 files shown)


src/Stoa.ts
```diff
@@ -21,11 +23,31 @@ class Stoa {
...
+         * JSON block data is parsed and stored on each storage.
+         */
         this.stoa.post("/push",
-            (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            console.log(req.body);
-            res.status(200).send();
-        });
+            asyncHandler(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+            var data: any = req.body;
+            var block: any = JSON.parse(data.block);
```
AndrejMitrovic: If parsing fails, what about sending 500 and logging the error?

src/Stoa.ts
```diff
@@ -21,11 +23,31 @@ class Stoa {
...
+            console.log(block);
+            var blockStorage: BlockStorage = await new BlockStorage(err);
+            await blockStorage.createTable(err);
+            await blockStorage.put(block, err);
```
AndrejMitrovic: This is creating a new BlockStorage class and then trying to create a table each time a block is received?

src/Stoa.ts
TrustHenry: Good catch! There is a possibility of an error.

src/Stoa.ts
TrustHenry: I think that's the problem.
I need to create a table from the constructor and create a generation status value.

TrustHenry: @MichaelKim20  is making improvements in this area.
About DB instances,

TrustHenry:
@AndrejMitrovic ping

src/modules/storage/BlockStorage.ts
```diff
@@ -81,23 +83,37 @@ export class BlockStorage extends Storages
             return;
         }
 
+        var tx_count: number = 0;
+        var enrollment_count: number = 0;
+
+        if (data.txs)
```
AndrejMitrovic: Transactions in a block are not optional. So why the `if` check here?

src/Stoa.ts
```diff
@@ -21,11 +23,41 @@ class Stoa {
...
-            (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            console.log(req.body);
-            res.status(200).send();
-        });
+            asyncHandler(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+            var data: any = req.body;
+            var block: any;
+
+            if (data.block == undefined)
+                res.status(500).send("Block data is not defined");
```
AndrejMitrovic: what happens after the if statement is true? We send back 500, and then? Does the delegate return? Or is the code missing return statements?

src/modules/storage/BlockStorage.ts
TrustHenry: That's right. The transactions is required.
Check the error above.

AndrejMitrovic: > Check the error above.

Will you add the check?

src/Stoa.ts
```diff
@@ -21,11 +23,41 @@ class Stoa {
             res.send(req.params.address + ':' + req.query.height);
         });
 
+        /**
+         * When a request is received through agora-history push
```
Geod24:
```suggestion
         * When a request is received through the `/push` handler
```

`agora-history` is just a tool for you to test.

src/Stoa.ts
```diff
@@ -21,11 +23,41 @@ class Stoa {
...
+            var data: any = req.body;
+            var block: any;
+
+            if (data.block == undefined)
+                res.status(500).send("Block data is not defined");
+
+            try {
+                block = JSON.parse(data.block);
+            } catch(e) {
+                res.status(500).send("Not a valid JSON format");
```
Geod24: It shouldn't be a 500, but a 400.

src/Stoa.ts
Geod24: Yes I don't quite understand that myself. Everything should be stored in the same database, and the schema should be initialized when the database is created.

Geod24 (changes requested):

src/modules/storage/BlockStorage.ts
TrustHenry: Check the block data.txs value. If not, return the error.

src/Stoa.ts
TrustHenry: Yes, I further subdivided the error-treatment logic.

src/Stoa.ts
TrustHenry: Apply the suggestion.

src/Stoa.ts
TrustHenry: I fixed it.

TrustHenry:
I deleted it "express-async-handler" 
This is difficult to subdivide errors in asynchronous processing.

TrustHenry:
> Yes I don't quite understand that myself. Everything should be stored in the same database, and the schema should be initialized when the database is created.

https://github.com/bpfkorea/stoa/pull/16
It's being addressed in that issue.

TrustHenry:
@Geod24 ping

src/Stoa.ts
```diff
@@ -21,11 +22,64 @@ class Stoa {
...
             (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            console.log(req.body);
-            res.status(200).send();
-        });
+            var data: any = req.body;
+            var block: any;
+
+            if (data.block == undefined)
+            {
+                res.status(400).send("Block data is not defined");
```
Geod24:
```suggestion
                res.status(400).send("Missing 'block' object in body");
```

Always tell the client what they are doing wrong, or even better, how to fix it. In this case we say what is wrong.

src/Stoa.ts
```diff
@@ -21,11 +22,64 @@ class Stoa {
...
+        /**
+         * When a request is received through the `/push` handler
+         * JSON block data is parsed and stored on each storage.
+         */
         this.stoa.post("/push",
             (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            console.log(req.body);
-            res.status(200).send();
-        });
+            var data: any = req.body;
```
Geod24: Can you just use `req.body` directly since it's only used twice ? [Follow the rule of three](https://blog.codinghorror.com/rule-of-three/).

src/Stoa.ts
```diff
@@ -21,11 +22,64 @@ class Stoa {
...
+            var block: any;
+
+            if (data.block == undefined)
+            {
+                res.status(400).send("Block data is not defined");
+                return;
+            }
+
+            try {
+                block = JSON.parse(data.block);
```
Geod24: If you access `req.body.block`... Isn't it already parsed ?

src/Stoa.ts
```diff
@@ -21,11 +22,64 @@ class Stoa {
...
+            } catch(e) {
+                res.status(400).send("Not a valid JSON format");
+                return;
+            }
+
+            console.log(block);
+            var blockStorage: BlockStorage = new BlockStorage((err: any) =>
```
Geod24: Why do you new the `BlockStorage` every time ? There should only be one needed.

src/Stoa.ts
```diff
@@ -21,11 +22,64 @@ class Stoa {
...
+            {
+                if (err != null)
+                {
+                    console.error(err);
+                    res.status(500).send("DB connection error");
+                    return;
+                }
+                blockStorage.createTable((err: any) =>
```
Geod24: Likewise, we should only need to initialize the schema at the beginning, why the `createTable` on every request??

src/modules/storage/BlockStorage.ts
```diff
@@ -43,11 +43,13 @@ export class BlockStorage extends Storages
...
-            validators TEXT,
-            merkle_root TEXT,
-            signature TEXT
+            height INTEGER NOT NULL PRIMARY KEY,
+            prev_block TEXT NOT NULL,
+            validators TEXT NOT NULL,
+            merkle_root TEXT NOT NULL,
+            signature TEXT,
+            tx_count INTEGER NOT NULL,
+            enrollment_count INTEGER NOT NULL DEFAULT 0
```
Geod24: Remove the default. Because our query will always insert a number, and the query is only done from one place. Hence default does not make sense.

src/modules/storage/BlockStorage.ts
```diff
@@ -74,30 +76,46 @@ export class BlockStorage extends Storages
                 (data.header.validators == undefined) ||
                 (data.header.validators._storage == undefined) ||
                 (data.header.merkle_root == undefined) ||
-                (data.header.signature == undefined)
+                (data.header.signature == undefined) ||
+                (data.txs == undefined)
         ) {
             if (callback != undefined)
                 callback("Parameter validation failed.");
```
Geod24: In a future PR, it would be good to say which parameter failed. And why it failed.

src/modules/storage/BlockStorage.ts
```diff
@@ -74,30 +76,46 @@ export class BlockStorage extends Storages
...
             return;
         }
 
+        var tx_count: number = 0;
+        var enrollment_count: number = 0;
+
+        tx_count = data.txs.length;
+
+        console.log(tx_count);
```
Geod24: Random `console.log` aren't great. Either remove or add an explanation message. In this case, I would say remove.

Geod24 (changes requested):

src/Stoa.ts
TrustHenry: Apply the suggetion.

src/Stoa.ts
TrustHenry: Delete Unnecessary Variables

src/Stoa.ts
TrustHenry: No, it's String.
It should be sent to application/json.

src/Stoa.ts
TrustHenry: This issue has been resolved in #16.

src/Stoa.ts
TrustHenry: This issue has been resolved in #16.

src/modules/storage/BlockStorage.ts
TrustHenry: This requires changing the table schema design first.

src/modules/storage/BlockStorage.ts
TrustHenry: In the future,  We'd better tell the customer the reason for the failure in detail.

src/modules/storage/BlockStorage.ts
TrustHenry: Removed it. This is a remnant of Debug.

TrustHenry:
@Geod24 ping

src/Stoa.ts
```diff
@@ -10,6 +11,17 @@ class Stoa {
         this.stoa.use(bodyParser.urlencoded({ extended: false }))
         // parse application/json
         this.stoa.use(bodyParser.json())
+        // DB fileName
+        var filename: string = ".cache";
```
Geod24: No, that shouldn't be the default.

src/Stoa.ts
```diff
@@ -10,6 +11,17 @@ class Stoa {
...
+        // create blockStorage
+        var blockStorage: BlockStorage = new BlockStorage(filename, (err: any) =>
+        {
+            if (err != null)
+            {
+                console.error("Failed to open database");
+                return;
```
Geod24: It's a simple return... So the code continues as if nothing happened ?

src/modules/storage/BlockStorage.ts
Geod24: Then let's do that. It's a very simple change.

src/Stoa.ts
```diff
@@ -21,10 +33,45 @@ class Stoa {
...
+            } catch(e) {
+                res.status(400).send("Not a valid JSON format");
+                return;
+            }
+
+            console.log(block);
+
+            blockStorage.put(block, (err: any) =>
+            {
+                blockStorage.close;
```
Geod24: Why `close` ??

src/Stoa.ts
```diff
@@ -21,10 +33,45 @@ class Stoa {
...
+                if (err != null)
+                {
+                    console.error(err);
+                    res.status(500).send("An error occurred while saving");
+                    return;
+                }
+                else
+                {
+                    console.log("200");
```
Geod24: This is useless

src/Stoa.ts
```diff
@@ -21,10 +33,45 @@ class Stoa {
...
+            }
+
+            console.log(block);
+
+            blockStorage.put(block, (err: any) =>
+            {
+                blockStorage.close;
+                if (err != null)
+                {
+                    console.error(err);
```
Geod24: This should come with a message so it can be grepped for

src/modules/storage/BlockStorage.ts
```diff
@@ -63,30 +65,44 @@ export class BlockStorage extends Storages
...
             return;
         }
 
+        var tx_count: number = 0;
+        var enrollment_count: number = 0;
+
+        tx_count = data.txs.length;
+
+        if (data.enrollments)
+            enrollment_count = data.enrollments.length;
```
Geod24:
```suggestion
        var tx_count: number = data.txs.length;
        var enrollment_count: number = 0;
        if (data.enrollments)
            enrollment_count = data.enrollments.length;
```

At the very least. And I believe you can do the same for enrollments ? Otherwise just use a ternary.

src/modules/storage/BlockStorage.ts
```diff
@@ -63,30 +65,44 @@ export class BlockStorage extends Storages
...
+            VALUES
+                (?, ?, ?, ?, ?, ?, ?)
+            ON CONFLICT(height) DO UPDATE SET
+                prev_block = prev_block,
+                validators = validators,
+                merkle_root = merkle_root,
+                signature = signature,
+                tx_count = tx_count,
+                enrollment_count = enrollment_count
+            `;
```
Geod24: There should **never** be a conflict. I'm not sure silent override is a good idea.

src/Stoa.ts
TrustHenry: Changed Input `Stor` ctor parameter.

src/Stoa.ts
TrustHenry: Stoa express will not be able to provide any service.
 No router URL is set up.

src/Stoa.ts
TrustHenry: Remove it.

src/Stoa.ts
TrustHenry: Yes

src/modules/storage/BlockStorage.ts
TrustHenry: used a ternary. (enrollments)

src/modules/storage/BlockStorage.ts
TrustHenry: Of course, INSERT should only happen once. 
BUT, There is currently no detailed error process.
If there is no `conflict`, it makes the current development and testing process difficult.
This can be eliminated if the error is subdivided.

src/Stoa.ts
TrustHenry: ? It returns an error message during storage.

TrustHenry:
@Geod24 ping

src/Stoa.ts
Geod24: The application should stop. Can this be made to throw ?

src/Stoa.ts
Geod24: It returns an unpredictable error message.
What I'm suggesting is:
```suggestion
                    console.error("Failed to store the payload of a push to the DB: " + err);
```

This way someone can grep for "Failed to store the payload of a push to the DB" and know where the error is logged. because `console.log(err)` is not something you can look for easily.

src/Stoa.ts
TrustHenry: I think we should make a exception .

src/main.ts
```diff
@@ -3,7 +3,7 @@ import Stoa from './Stoa';
 import express from "express";
 
 const port: number = Number(process.env.PORT) || 3836;
-const stoa: express.Application = new Stoa().stoa;
+const stoa: express.Application = new Stoa(".cache").stoa;
```
Geod24:
```suggestion
const stoa: express.Application = new Stoa("database").stoa;
```

Don't use `.cache`, that's what Agora do, and at the very least don't do the same thing as Agora to avoid conflict if someone inadvertently runs agora AND stoa in the same working directory. It shouldn't be the case, but if it happens, let's remove the obvious pitfalls.

src/modules/storage/BlockStorage.ts
Geod24: My comment means: Better to show the error than to hide it. It is okay not to handle the error now. But if the error happens, you want to know it. If we know it, we can fix it easily. If we hide it like this, we might loose time looking for it.

src/main.ts
TrustHenry: I apply the suggestion.

src/modules/storage/BlockStorage.ts
TrustHenry: If so, I will remove the `conflict` part.

TrustHenry:
@Geod24 ping

The name changes from blockStorage to ledgerStorage and all DB query operations are processed here.
The type has been applied to error handling.

However, due to many changes, the Commit Separation task for the name change has been combined.

But if you don't mind, I think we can proceed with this as it is.
This is an obstacle to our next works.

TrustHenry:
Also, in the next PR, enrollmentStorage will be deleted and integrated into the ledgerStorage.

Geod24 (approved):

PR #16, opened 2020-06-23
Change reuse of one storage instance multiple times

MichaelKim20:
Remove the opening database from the constructor
Add the open method to Storage.
Enable testing with memory database

### DIFF
--- a/src/modules/storage/Storages.ts
+++ b/src/modules/storage/Storages.ts
@@ -18,7 +18,7 @@ import * as sqlite from 'sqlite3';
 export class Storages
 {
     /**
-     *  The instance of sqlite3
+     *  The instance of sqlite
      */
     protected db: sqlite.Database;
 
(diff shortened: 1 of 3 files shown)


src/modules/storage/Storages.ts
```diff
@@ -20,27 +20,32 @@ export class Storages
     /**
      *  The instance of sqlite3
      */
-    protected db: sqlite.Database;
+    protected db: any;
```
TrustHenry: Why any?  `sqlite.Database`

src/modules/storage/Storages.ts
```diff
@@ -20,27 +20,32 @@ export class Storages
...
-        let filename: string = ".cache";
-        this.db = new sqlite.Database(filename,
+        this.db = new sqlite.Database(this.db_path,
             sqlite.OPEN_CREATE | sqlite.OPEN_READWRITE |
-            sqlite.OPEN_SHAREDCACHE, (err: any) =>
-            {
-                if (callback != undefined)
-                    console.error(err);
-            });
+            sqlite.OPEN_SHAREDCACHE);
```
TrustHenry: Error handling is missing when opening database.

src/modules/storage/Storages.ts
```diff
@@ -20,27 +20,32 @@ export class Storages
...
+    }
+
+    /**
+     * Open the database
+     */
+    public open ()
     {
-        let filename: string = ".cache";
-        this.db = new sqlite.Database(filename,
+        this.db = new sqlite.Database(this.db_path,
```
TrustHenry: What if we create it only when we have a db instance?

src/modules/storage/Storages.ts
MichaelKim20: I made the type 'any' because if the member variable is not generated by the constructor, the error occurs.

Geod24: That's not an improvement.

Geod24:
It looks like `Storages` and `BlockStorages` are redundant, aren't they ?

MichaelKim20:
> It looks like `Storages` and `BlockStorages` are redundant, aren't they ?

We are currently inheriting Storage to create EnrollmentStorage and TransactionStorage.

src/modules/storage/Storages.ts
MichaelKim20: Do you want this? Please describe the points more specifically.

```
        if (this.db != null)
            this.db.close();
        this.db = new sqlite.Database(this.db_path,
            sqlite.OPEN_CREATE | sqlite.OPEN_READWRITE |
            sqlite.OPEN_SHAREDCACHE);
        this.db.configure("busyTimeout", 1000);
```

src/modules/storage/Storages.ts
MichaelKim20: It's not an error handle, it's just a callback function.
This is when callback is executed.db is not set yet.
Therefore, I removed it because I thought it was difficult to use it correctly.

src/modules/storage/Storages.ts
MichaelKim20: I change to `protected db: sqlite.Database|any;`

TrustHenry (approved):
LGTM

TrustHenry (approved):
LGTM

MichaelKim20:
@Geod24 ping

Geod24 (approved):

PR #18, opened 2020-06-24
Add the class and test code to process transactions data

MichaelKim20:
The ability to store all transaction data in the block in the database has been implemented.
I implemented it according to what was defined in the database schema.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -14,6 +14,8 @@
 
 *******************************************************************************/
 
+import * as sqlite from 'sqlite3';
+import { Hash }  from '../common/Hash'
 import { Storages } from './Storages';
 
 /**
@@ -21,6 +23,14 @@ import { Storages } from './Storages';
  */
 export class LedgerStorage extends Storages
 {
+    private hash: Hash;
+
+    constructor (filename: string, callback: (err: Error | null) => void)
+    {
+        super(filename, callback);
+        this.hash = new Hash();
+    }
+
     /**
      * Creates tables related to the ledger.
      * @param callback If provided, this function will be called when
@@ -29,7 +39,7 @@ export class LedgerStorage extends Storages
      */
     public createTables (callback: (err: Error | null) => void)
     {
-        var sql =
+        var sql: string =
         `CREATE TABLE IF NOT EXISTS blocks
         (
             height INTEGER NOT NULL PRIMARY KEY,
@@ -50,11 +60,46 @@ export class LedgerStorage extends Storages
             cycle_length INTEGER NOT NULL,
             enroll_sig TEXT NOT NULL,
             PRIMARY KEY("block_height","enrollment_index")
-        );`;
-        this.db.each(sql, (err: Error | null) =>
+        );
+
+        CREATE TABLE IF NOT EXISTS "transactions"
+        (
+            "block_height"      INTEGER NOT NULL,
+            "tx_index"          INTEGER NOT NULL,
+            "tx_hash"           TEXT NOT NULL,
+            "type"              INTEGER NOT NULL,
+            "inputs_count"      INTEGER NOT NULL,
+            "outputs_count"     INTEGER NOT NULL,
+            PRIMARY KEY("block_height", "tx_index")
+        );
+
+        CREATE TABLE IF NOT EXISTS "tx_inputs"
+        (
+            "block_height"      INTEGER NOT NULL,
+            "tx_index"          INTEGER NOT NULL,
+            "in_index"          INTEGER NOT NULL,
+            "previous"          TEXT NOT NULL,
+            "out_index"         INTEGER NOT NULL,
+            PRIMARY KEY("block_height", "tx_index", "in_index")
+        );
+
+        CREATE TABLE IF NOT EXISTS "tx_outputs"
+        (
+            "block_height"  INTEGER NOT NULL,
+            "tx_index"      INTEGER NOT NULL,
+            "output_index"  INTEGER NOT NULL,
+            "tx_hash"       TEXT NOT NULL,
+            "utxo_key"      TEXT NOT NULL,
+            "amount"        NUMERIC NOT NULL,
+            "address"       TEXT NOT NULL,
+            "used"          INTEGER NOT NULL DEFAULT 0,
+            PRIMARY KEY("block_height", "tx_index", "output_index")
+        );
+        `;
+
+        this.db.exec(sql, (err: Error | null) =>
         {
-            if (callback != null)
-                callback(err);
+            callback(err);
         });
     }
 
@@ -197,7 +242,7 @@ export class LedgerStorage extends Storages
 
             this.putEnrollment(enrollment, (err: Error | null) =>
             {
-                if (err != null)
+                if (!err)
                 {
                     idx++;
                     doPut();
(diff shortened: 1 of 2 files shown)


src/modules/storage/TransactionsStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+            "type"              INTEGER NOT NULL,
+            "inputs_count"      INTEGER NOT NULL,
+            "outputs_count"     INTEGER NOT NULL,
+            PRIMARY KEY("block_height", "tx_index")
+        )`;
+        this.db.run(sql, (err: any) =>
+        {
+            if (callback != undefined)
+                callback(err);
+        });
```
Geod24: This separation is useless. `createTable` doesn't do anything more than `createTransactionsTable`. Merge them.

src/modules/storage/TransactionsStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+
+*******************************************************************************/
+
+import * as sqlite from 'sqlite3';
+import { Storages } from './Storages';
+
+/**
+ * The class that insert and read the transactions into the database.
+ */
+export class TransactionsStorage extends Storages
```
Geod24: Why do we need to have those `Storages` class ? We have a single DB where we store everything. Why not just a single `Database` class that does the work ? We don't put single transactions at the moment.

src/modules/storage/TransactionsStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+                (tx_index == undefined) ||
+                (tx_hash == undefined)
+        ) {
+            if (callback != undefined)
+                callback("Parameter validation failed.");
+            return;
+        }
+
+        let sql =
+        `INSERT OR REPLACE INTO transactions
```
Geod24: Silent override is a terrible default, don't do that.

src/modules/storage/TransactionsStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+    {
+        var sql =
+        `SELECT
+            block_height, tx_index, tx_hash, type, inputs_count, outputs_count
+        FROM
+            transactions
+        WHERE block_height = ?`;
+        this.db.all(sql, [height], (err: any, rows: any) =>
+        {
+            if (callback != undefined)
```
Geod24: It doesn't make sense to provide an undefined callback here, because that means `getTransactions` has no effect.

src/modules/storage/TransactionsStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Gets a transaction data
+     * @param height the height of the block to get
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     * The second argument is result set.
+     */
+    public getTransactions (height: any, callback?: any)
```
Geod24: Please stop using `any` for the callback and give it a proper type.

Geod24 (changes requested):
- This PR does much more than it pretends to. The first two commit in particular.
- There are a few issues with the way things are approached;
- Please stop using `any` everywhere;

MichaelKim20:
Ready for review

src/modules/storage/LedgerStorage.ts
```diff
@@ -14,20 +14,30 @@
...
+    }
+
     /**
      * Creates tables related to the ledger.
      * @param callback If provided, this function will be called when
      * the database was finished successfully or when an error occurred.
      * The first argument is an error object.
      */
-    public createTables (callback: (err: Error | null) => void)
+    public createTable (callback?: (err: Error | null) => void)
```
```suggestion
    public createTables (callback: (err: Error | null) => void)
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -119,4 +164,380 @@ export class LedgerStorage extends Storages
...
+     * @param data a transaction data
+     * @param callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     * The first argument is an error object.
+     */
+    private putTransaction (block_height: number, tx: any, tx_index: number,
+        tx_hash: string, callback?: (err: Error | null) => void)
+    {
+        if (tx == undefined) {
+            if (callback != undefined)
```
TrustHenry, line 390:
```suggestion
            if (callback != null)
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -119,4 +164,380 @@ export class LedgerStorage extends Storages
...
+            if (err)
+            {
+                if (callback != undefined)
+                    callback(err);
+                return;
+            }
+
+            this.putInputs(block_height, tx, tx_index, tx_hash, (err2: any) =>
+            {
+                if (err2)
```
TrustHenry, line 419:
```suggestion
                if (err2 != null)
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -119,4 +164,380 @@ export class LedgerStorage extends Storages
...
+            ], (err: Error | null) =>
+        {
+            if (err)
+            {
+                if (callback != undefined)
+                    callback(err);
+                return;
+            }
+
+            this.putInputs(block_height, tx, tx_index, tx_hash, (err2: any) =>
```
```suggestion
            this.putInputs(block_height, tx, tx_index, tx_hash, (err2: ERROR | null) =>
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -119,4 +164,380 @@ export class LedgerStorage extends Storages
...
+            {
+                if (err2)
+                {
+                    if (callback != undefined)
+                        callback(err2);
+                    return;
+                }
+
+                this.putOutputs(block_height, tx, tx_index, tx_hash, (err3: any) =>
```
```suggestion
                this.putOutputs(block_height, tx, tx_index, tx_hash, (err3: ERROR | null) =>
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -39,7 +49,42 @@ export class LedgerStorage extends Storages
...
+        CREATE TABLE IF NOT EXISTS "tx_outputs"
+        (
+            "block_height"  INTEGER NOT NULL,
+            "tx_index"      INTEGER NOT NULL,
+            "output_index"  INTEGER NOT NULL,
+            "tx_hash"       TEXT NOT NULL,
+            "utxo_key"      TEXT NOT NULL,
+            "amount"        NUMERIC NOT NULL,
+            "address"       TEXT(56) NOT NULL,
+            "used"          TEXT(1) NOT NULL DEFAULT 'N',
```
AndrejMitrovic: Why is this `TEXT` instead of just a bool or an integer? Also SQLite ignores these limits, it's not like in MySQL where it's a hard limit.

src/modules/storage/LedgerStorage.ts
MichaelKim20: If no parameter is specified, it is `undefined`. Therefore, there is no need to modify it.

src/modules/storage/LedgerStorage.ts
AndrejMitrovic: Sorry there is no bool, but it should be an integer.

MichaelKim20: Thanks. I will fix

src/modules/storage/LedgerStorage.ts
```diff
@@ -14,20 +14,65 @@
...
+                        return;
+                    }
+
+                    this.createOutputsTable((err4: Error | null) =>
+                    {
+                        callback(err4);
+                    });
+                });
+            });
+        });
```
Geod24: Why the split ? This makes the code worse for no reason.

MichaelKim20: I used another method, and it didn't work.
I'll have to do it this way.

MichaelKim20: I handled this in the way @TrustHenry  applied.

src/modules/storage/TransactionsStorage.ts
MichaelKim20: I changed

src/modules/storage/TransactionsStorage.ts
MichaelKim20: I changed

src/modules/storage/TransactionsStorage.ts
MichaelKim20: I changed

src/modules/storage/LedgerStorage.ts
MichaelKim20: I changed

src/modules/storage/LedgerStorage.ts
MichaelKim20: if (!err) "if not error..." This expression is more intuitive here. So I prefer this.

src/modules/storage/LedgerStorage.ts
MichaelKim20: I changed, Thanks

src/modules/storage/LedgerStorage.ts
MichaelKim20: I changed, Thanks

src/modules/storage/LedgerStorage.ts
MichaelKim20: I changed, Thanks
``` SQL
            "address"       TEXT NOT NULL,
            "used"          INTEGER NOT NULL DEFAULT 0,
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -235,4 +348,403 @@ export class LedgerStorage extends Storages
             callback(err, rows);
         });
     }
+
+    /**
+     * Get enrollments
```
TrustHenry: Invalid getValidator function and description.

src/modules/storage/LedgerStorage.ts
MichaelKim20: I changed. Thanks.
``` TypeScript

    public createTables (callback: (err: Error | null) => void)
    {
        var sql: string =
        `CREATE TABLE IF NOT EXISTS blocks
        (
            height INTEGER NOT NULL PRIMARY KEY,
            prev_block TEXT NOT NULL,
            validators TEXT NOT NULL,
            merkle_root TEXT NOT NULL,
            signature TEXT,
            tx_count INTEGER NOT NULL,
            enrollment_count INTEGER NOT NULL
        );

        CREATE TABLE IF NOT EXISTS enrollments
        (
            block_height INTEGER NOT NULL,
            enrollment_index INTEGER NOT NULL,
            utxo_key TEXT NOT NULL,
            random_seed TEXT NOT NULL,
            cycle_length INTEGER NOT NULL,
            enroll_sig TEXT NOT NULL,
            PRIMARY KEY("block_height","enrollment_index")
        );

        CREATE TABLE IF NOT EXISTS "transactions"
        (
            "block_height"      INTEGER NOT NULL,
            "tx_index"          INTEGER NOT NULL,
            "tx_hash"           TEXT NOT NULL,
            "type"              INTEGER NOT NULL,
            "inputs_count"      INTEGER NOT NULL,
            "outputs_count"     INTEGER NOT NULL,
            PRIMARY KEY("block_height", "tx_index")
        );

        CREATE TABLE IF NOT EXISTS "tx_inputs"
        (
            "block_height"      INTEGER NOT NULL,
            "tx_index"          INTEGER NOT NULL,
            "in_index"          INTEGER NOT NULL,
            "previous"          TEXT NOT NULL,
            "out_index"         INTEGER NOT NULL,
            PRIMARY KEY("block_height", "tx_index", "in_index")
        );

        CREATE TABLE IF NOT EXISTS "tx_outputs"
        (
            "block_height"  INTEGER NOT NULL,
            "tx_index"      INTEGER NOT NULL,
            "output_index"  INTEGER NOT NULL,
            "tx_hash"       TEXT NOT NULL,
            "utxo_key"      TEXT NOT NULL,
            "amount"        NUMERIC NOT NULL,
            "address"       TEXT NOT NULL,
            "used"          INTEGER NOT NULL DEFAULT 0,
            PRIMARY KEY("block_height", "tx_index", "output_index")
        );

        CREATE TABLE IF NOT EXISTS "validators"
        (
            "enrolled_at"	INTEGER NOT NULL,
            "utxo_key"	TEXT NOT NULL,
            "address"	TEXT(56) NOT NULL,
            "amount"	NUMERIC NOT NULL,
            "preimage_distance"	INTEGER NOT NULL,
            "preimage_hash"	TEXT NOT NULL,
            PRIMARY KEY("enrolled_at","utxo_key")
        );
        `;

        this.db.exec(sql, (err: Error | null) =>
        {
            callback(err);
        });
    }
```

tests/Storage.test.ts
```diff
@@ -292,43 +292,61 @@ var sample_data =
...
+    {
+        assert.ok(!err2, err2?.message);
+
+        getBlockData(ledger_storage, 1, (err3: Error | null) =>
+        {
+            assert.ok(!err3, err3?.message);
 
-                ledger_storage.close();
-              });
+            callback();
```
TrustHenry: The test code for putAllBlockData is written.
However, it is not actually connected to the Stoa API.
https://github.com/bpfkorea/stoa/blob/5b3252c8abb9a215ea7c4b8903fc8cb2111a2269/src/Stoa.ts#L59
You need to change the code of this Stoa to associate block data storage with the API.

MichaelKim20: Good Point

MichaelKim20: I connected another methods to the end of LedgerStorage.putBlocks

src/modules/storage/LedgerStorage.ts
MichaelKim20: I fixed it. thanks

TrustHenry (approved):
LGTM

src/modules/storage/LedgerStorage.ts
Geod24: So do I understand correctly we have three states:
- undefined
- null
- valid

And the undefined comes from the `?` ? In which case either `?` is redundant, or `| null` is, isn't it ?

Geod24: Ah nevermind, it's the parameter that can be `null`, not the callback itself

src/modules/storage/LedgerStorage.ts
```diff
@@ -235,4 +280,381 @@ export class LedgerStorage extends Storages
...
+                    {
+                        if (callback != undefined)
+                            callback(err);
+                        else
+                            return;
+                    }
+                }
+            );
+        }
+        doPut();
```
Geod24, line 472: This is a very convoluted pattern I've seen a few times. It was mentioned that it was for asynchronicity ?
I don't quite follow how this helps, there's no `async` or `await` here.

MichaelKim20: I looked for it for a long time because I wanted to handle it like that.
I haven't found a good solution yet.
Since I need to process multiple queries sequentially, I have to get a callback and process them.
I think I can use a function called serial, but we need more time.

PR #19, opened 2020-06-24
Apply the new database schema in the `BlockStorage`

MichaelKim20:
Apply the new database schema in the `BlockStorage`


MichaelKim20:
I will close this because the related content is in Henry's PR. #14

PR #20, opened 2020-06-24
Remove the enrollmentStorage and integrate with LedgerStorage

TrustHenry:
DB Instance is managed in one class and 
the required query operations are managed in LedgerStorage.


src/modules/storage/LedgerStorage.ts
```diff
@@ -119,4 +130,107 @@ export class LedgerStorage extends Storages
...
+                {
+                    if (callback != null)
+                        callback(err);
+                    else
+                        return;
+                }
+            });
+        }
+        doPut();
+    }
```
Geod24, line 213: Why does a foreach have to be 30 lines ?

Geod24: Ah you are just moving the code

Geod24 (approved):

src/modules/storage/LedgerStorage.ts
AndrejMitrovic: More importantly, why is this recursive?

TrustHenry: You're watching a feast of Asynchronous Bracelet.
If I find and write a good asynchronous library, I think the code will be simple.
I don't have enough time to apply it.

AndrejMitrovic: I think this is the wrong place to write asynchronous code.

TrustHenry: @AndrejMitrovic 
I didn't understand what you meant.

PR #21, opened 2020-06-24
Fix a bug that does not run each query on the Create table

TrustHenry:
Fix the error handling callback that cannot be caught.


MichaelKim20 (approved):
LGTM

Issue #9, opened 2020-06-22
Implement shell API server with `/push`, `/validators` and `/validators/{address}` endpoints


Geod24:
# Abstract

API to implement:
```
POST /push
GET /validators[?height=42]
GET /validator/{address}[?height=42]
```

## Push

The `/push` endpoint accepts and stores a block in a relational database. The schema of the database has to be documented (#7 ). No validation is required on the incoming data.
A tool to test this is implemented in bpfkora/agora#972

## Validators

The `height` parameter is optional. If present, it must accepts a positive integer. If absent, the most up to date state is expected.
`{address}` is the public key string of a node.

The return of `GET /validators` is an array of `ValidatorInfo` items, JSON serialized.
```json
[
    {
        "address": "GCEEEEEEE424242",
        "enrolled_at": 4242,
        "stake": "0xabcdefdeadbeef",
        "preimage": {
            "distance": 69,
            "hash": "0xdeadbeefbadf00d"
        }
    }
]
```

TrustHenry:
I think the `stake` value here is the UTXO amount of the validator.

Geod24:
It's the UTXO key, not amount.

PR #22, opened 2020-06-24
New endpoint Development API GET '/validators'

TrustHenry:
It provides two endpoints.
GET /validators
GET /validator/{address}
a height query parameter is an option.
Look up the validators stored in the leaderStorage by conditions and respond with JSON data.

Require #18 

Relates #9

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -1,6 +1,7 @@
 import express from "express";
 import bodyParser from "body-parser";
 import { LedgerStorage } from "./modules/storage/LedgerStorage";
+import { ValidatorData, IPreimage, IValidator } from "./modules/data/ValidatorData";
 
 class Stoa {
     public stoa: express.Application;
@@ -23,14 +24,98 @@ class Stoa {
             }
         });
 
+        /**
+         * Called when a request is received through the `/validators` handler
+         *
+         * Returns a set of Validators based on the block height if there is a height.
+         * If height was not provided the latest validator set is returned.
+         */
         this.stoa.get("/validators",
             (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            res.send(req.query.height);
+
+            var height: number = Number(req.query.height);
+
+            if (!Number.isNaN(height) && (!Number.isInteger(height) || height < 0))
+            {
+                res.status(400).send("The Height value is not valid.");
+                return;
+            }
+
+            this.ledger_storage.getValidatorsAPI(height, null, (err: Error | null, rows: any[]) =>
+            {
+                if (err != null)
+                {
+                    console.error("Failed to data lookup to the DB: " + err);
+                    res.status(500).send("Failed to data lookup");
+                    return;
+                }
+
+                if (rows.length)
+                {
+                    let out_put:Array<ValidatorData> = new Array<ValidatorData>();
+
+                    for (const row of rows)
+                    {
+                        let preimage: IPreimage = {distance: row.distance, hash: '0'} as IPreimage;
+                        var validator: ValidatorData =
+                            new ValidatorData(row.address, row.enrolled_at, row.stake, preimage);
+                        out_put.push(validator);
+                    }
+                    res.status(200).send(JSON.stringify(out_put));
+                }
+                else
+                {
+                    res.status(204).send();
+                }
+            });
         });
 
+        /**
+         * Called when a request is received through the `/validators/:address` handler
+         *
+         * Returns a set of Validators based on the block height if there is a height.
+         * If height was not provided the latest validator set is returned.
+         * If an address was provided, return the validator data of the address if it exists.
+         */
         this.stoa.get("/validator/:address",
             (req: express.Request, res: express.Response, next: express.NextFunction) => {
-            res.send(req.params.address + ':' + req.query.height);
+
+            var height: number = Number(req.query.height);
+            var address: string = String(req.params.address);
+            
+            if (!Number.isNaN(height) && (!Number.isInteger(height) || height < 0))
+            {
+                res.status(400).send("The Height value is not valid.");
+                return;
+            }
+
+            this.ledger_storage.getValidatorsAPI(height, address, (err: Error | null, rows: any[]) =>
+            {
+                if (err != null)
+                {
+                    console.error("Failed to data lookup to the DB: " + err);
+                    res.status(500).send("Failed to data lookup");
+                    return;
+                }
+
+                if (rows.length)
+                {
+                    let out_put:Array<ValidatorData> = new Array<ValidatorData>();
+
+                    for (const row of rows)
+                    {
+                        let preimage: IPreimage = {distance: row.distance, hash: '0'} as IPreimage;
+                        var validator: ValidatorData | null =
+                            new ValidatorData(row.address, row.enrolled_at, row.stake, preimage);
+                        out_put.push(validator);
+                    }
+                    res.status(200).send(JSON.stringify(out_put));
+                }
+                else
+                {
+                    res.status(204).send();
+                }
+            });
         });
 
         /**
--- a/src/modules/data/ValidatorData.ts
+++ b/src/modules/data/ValidatorData.ts
@@ -1,0 +1,25 @@
+export interface IValidator {
+    address: string;
+    enrolled_at: number;
+    stake: string;
+    preimage?: IPreimage;
+};
+
+export interface IPreimage {
+    distance?: number;
+    hash?: string;
+};
+
+export class ValidatorData implements IValidator {
+    address: string;
+    enrolled_at: number;
+    stake: string;
+    preimage: IPreimage;
+
+    constructor(address: string, enrolled_at: number, stake: string, preimage: IPreimage) {
+        this.address = address;
+        this.enrolled_at = enrolled_at;
+        this.stake = stake;
+        this.preimage = preimage;
+    }
+}
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -752,4 +752,47 @@ export class LedgerStorage extends Storages
             callback(err, rows);
         });
     }
+
+    /**
+     * Get validators
+     * @param height: The height parameter is optional.
+     * validators based on the block height if there is a height.
+     * if height null the most up to date state is expected.
+     * if the height is null then current valid validators.
+     * @param Callback If provided, this function will be called when
+     * the database was finished successfully or when an error occurred.
+     */
+    public getValidatorsAPI (height: number, address: string | null,
+        callback: (err: Error | null, rows: any[]) => void)
+    {
+        var cur_height: string;
+
+        if (!Number.isNaN(height))
+            cur_height = height.toString();
+        else
+            cur_height = `(SELECT MAX(height) as height FROM blocks)`;
+
+        var sql =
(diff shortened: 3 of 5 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -235,4 +236,44 @@ export class LedgerStorage extends Storages
...
+        `;
+        if (address != null)
+            sql += "and address = " + address;
+
+        var cur_height: string;
+
+        if (height != null)
+            cur_height = height.toString();
+        else
+            cur_height = `(SELECT MAX(height) as height FROM blocks)`;
```
TrustHenry: It is better to apply max height after query separately.

src/modules/storage/LedgerStorage.ts
```diff
@@ -235,4 +236,44 @@ export class LedgerStorage extends Storages
...
+    {
+        var sql =
+        `SELECT tx_outputs.address,
+                enrollments.block_height as enrolled_at,
+                enrollments.utxo_key as stake,
+                (? - (enrollments.block_height + 1)) as distance,
+                enrollments.random_seed
+        FROM enrollments
+            LEFT JOIN tx_outputs ON enrollments.utxo_key = tx_outputs.utxo_key
+        WHERE 1=1
```
Geod24: ?

src/modules/storage/LedgerStorage.ts
```diff
@@ -235,4 +236,44 @@ export class LedgerStorage extends Storages
...
+            and enrollments.block_height >= (? - enrollments.cycle_length)
+            and enrollments.block_height <= ?
```
Geod24: Since you used uppercase for other SQL keywords, let's be consistent and use `AND` here.

src/modules/data/ValidatorData.ts
```diff
@@ -0,0 +1,30 @@
+export interface IValidator {
+    address: string;
+    enrolled_at: number;
+    stake: string;
+    preimage?: Array<IPreimage>;
```
Geod24: Array ?

Geod24 (commented):
Nice!

src/modules/data/ValidatorData.ts
TrustHenry: Well, is it a single object?

TrustHenry: Oh, I knew JSON sample didn't have '[]'

Geod24: It is a single object, yes.

src/modules/storage/LedgerStorage.ts
TrustHenry: It is a meaningless code for readability when working on dynamic queries.
I delete it because it is my old coding habit.

src/Stoa.ts
```diff
@@ -23,14 +24,85 @@ class Stoa {
...
+                    }
+
+                    if (rows.length)
+                    {
+                        let out_put:Array<ValidatorData> = new Array<ValidatorData>();
+
+                        for (const row of rows)
+                        {
+                            let preimage: IPreimage = {distance: row.distance, hash: '0'} as IPreimage;
+                            var validator: ValidatorData | null =
```
AndrejMitrovic: What does `| null` do here?

src/Stoa.ts
```diff
@@ -23,14 +24,85 @@ class Stoa {
...
+                                new ValidatorData(row.address, row.enrolled_at, row.stake, preimage);
+                            out_put.push(validator);
+                        }
+                        res.status(200).send(JSON.stringify(out_put));
+                    }
+                    else
+                    {
+                        res.status(200).send("no data");
```
AndrejMitrovic: Actually this should never happen. There must always be validators active at any block height.

src/Stoa.ts
```diff
@@ -23,14 +24,85 @@ class Stoa {
...
+                    }
+                });
         });
 
+        /**
```
AndrejMitrovic, line 73: Is this documentation javascript-style convention? Because it's written in a very weird way. Where are the verbs describing the actions?

When.. => Called when..
Validators based on => Returns a set of Validators based on..
If height null the most up to date state is expected. => If height was not provided the latest validator set is returned
If an address exists => If an address was provided, return the validator data of the address if it exists

src/Stoa.ts
```diff
@@ -23,14 +24,85 @@ class Stoa {
...
+                        let preimage: IPreimage = {distance: row.distance, hash: '0'} as IPreimage;
+                        var validator: ValidatorData | null =
+                            new ValidatorData(row.address, row.enrolled_at, row.stake, preimage);
+                        out_put.push(validator);
+                    }
+                    res.status(200).send(JSON.stringify(out_put));
+                }
+                else
+                {
+                    res.status(200).send("no data");
```
AndrejMitrovic: Does it really make sense that we return 200 and a random string as the result? When designing APIs think about what the client code will look like. You will end up having to check for this string.

src/modules/storage/LedgerStorage.ts
```diff
@@ -50,11 +60,57 @@ export class LedgerStorage extends Storages
...
+            "amount"        NUMERIC NOT NULL,
+            "address"       TEXT NOT NULL,
+            "used"          INTEGER NOT NULL DEFAULT 0,
+            PRIMARY KEY("block_height", "tx_index", "output_index")
+        );
+
+        CREATE TABLE IF NOT EXISTS "validators"
+        (
+            "enrolled_at"	INTEGER NOT NULL,
+            "utxo_key"	TEXT NOT NULL,
```
AndrejMitrovic: There is a mix of spaces and tabs here. I think there is a setting in vscode so it never accidentally adds tabs.

src/modules/storage/LedgerStorage.ts
```diff
@@ -747,4 +747,45 @@ export class LedgerStorage extends Storages
...
+        `SELECT tx_outputs.address,
+                enrollments.block_height as enrolled_at,
+                enrollments.utxo_key as stake,
+                (` + cur_height + ` - (enrollments.block_height + 1)) as distance,
+                enrollments.random_seed,
+                (SELECT MAX(height) as height FROM blocks) as height,
+                ((SELECT MAX(height) as height FROM blocks) - (enrollments.block_height + 1)) as test
+        FROM enrollments
+            LEFT JOIN tx_outputs ON enrollments.utxo_key = tx_outputs.utxo_key
+        WHERE AND enrollments.block_height >= (` + cur_height + ` - enrollments.cycle_length)
```
```suggestion
        WHERE enrollments.block_height >= (` + cur_height + ` - enrollments.cycle_length)
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -747,4 +747,45 @@ export class LedgerStorage extends Storages
...
+            AND enrollments.block_height <= ` + cur_height + `
+        `;
+
+        if (address != null)
+            sql += `AND tx_outputs.address = '` + address + `'`;
```
```suggestion
            sql += ` AND tx_outputs.address = '` + address + `'`;
```

tests/Storage.test.ts
```diff
@@ -520,6 +523,35 @@ function putAllTransactionData (ledger_storage: LedgerStorage,
     doPut();
 }
 
+/**
+ * Run Validators API tests
+ */
+function runValidatorsAPITest (ledger_storage: LedgerStorage, callback?: (err: Error | null, rows: any[]) => void)
```
MichaelKim20:
```suggestion
function runValidatorsAPITest (ledger_storage: LedgerStorage, callback: () => void)
```

tests/Storage.test.ts
```diff
@@ -520,6 +523,35 @@ function putAllTransactionData (ledger_storage: LedgerStorage,
...
+          assert.ok(!err2, err2?.message);
+          assert.equal(rows.length, 1);
+          assert.equal(rows[0].address, address);
+
+          var res2 = ledger_storage.getValidatorsAPI(null, null, (err3: Error | null, rows: any[]) =>
+          {
+              assert.ok(!err3, err3?.message);
+              assert.equal(rows.length, 3);
+              assert.equal(rows[0].distance, 0);
+          });
```
MichaelKim20: Callback should be called after the test as there may be other tests.
```suggestion
              assert.equal(rows[0].distance, 0);
              callback();
          });
```

src/Stoa.ts
TrustHenry: Good catch! Null is not possible here.

src/Stoa.ts
TrustHenry: Yes, that's right. It won't happen, but if the height of the lookup condition exceeds the current block height, there may be no data. I will change.`res.status(200).send();`

AndrejMitrovic: Then it should return some kind of status code so the client knows it called with invalid parameters. For example `204` like Mathias suggested: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

tests/Storage.test.ts
```diff
@@ -360,46 +382,180 @@ function getBlockData (ledger_storage: LedgerStorage, height: any, callback?: an
...
+      assert.ok(!err1, err1?.message);
+      assert.equal(rows[0].address, address);
+      assert.equal(rows[0].enrolled_at, 0);
+      assert.equal(rows[0].distance, 0);
+
+      var res1 = ledger_storage.getValidatorsAPI(1, address, (err2: Error | null, rows: any[]) =>
+      {
+          assert.ok(!err2, err2?.message);
+          assert.equal(rows.length, 1);
+          assert.equal(rows[0].address, address);
```
MichaelKim20, line 474:
```suggestion
          assert.equal(rows[0].address, address);
          assert.equal(rows[0].stake, '0x210b66053c73e7bd7b27673706f0272617d09b8cda76605e91ab66ad'+
          '1cc3bfc1f3f5fede91fd74bb2d2073de587c6ee495cfb0d981f03a83651b48ce0e576a1a');
```
I'd like you to check on `stake`, too.

src/Stoa.ts
TrustHenry: Thank you.

TrustHenry: I applied JSDoc style.

src/Stoa.ts
TrustHenry: Right. I change the code now `res.status(204).send();`

src/modules/storage/LedgerStorage.ts
TrustHenry: I applied it now

src/modules/storage/LedgerStorage.ts
TrustHenry: Good catch!

tests/Storage.test.ts
TrustHenry: @MichaelKim20 Thanks you

TrustHenry:
@Geod24 @AndrejMitrovic @MichaelKim20 Thanks for reviews.
All suggestions have been applied.

src/modules/storage/LedgerStorage.ts
MichaelKim20: I fixed it

TrustHenry:
I'm doing a rebase.

TrustHenry:
@Geod24 ping 
Rebase is complete.

MichaelKim20 (approved):
LGTM

TrustHenry:
@Geod24 ping

PR #23, opened 2020-06-26
Remove  `any` type by defining the class associated with the block

MichaelKim20:
Remove `any` type by defining the class associated with the block

This allows us to know what changed when Agora's block structure changed.

PR #24, opened 2020-06-26
Change the recursive pattern to the simple `Array.forEach`

MichaelKim20:
I changed the recursive pattern to the simple repeating statement using 'Array.forEach' and 'sqlite.Statement'.

This requires PR #23


MichaelKim20:
I integrated this into PR #25

PR #25, opened 2020-06-26
Separate the callback function into two (success, error).

MichaelKim20:
This is the work of separating `callback` into `success` and `error`.

The code before work
```
    public putBlocks (data: any, callback?: (err: Error | null) => void)
    {
        ...
    }

    public getBlocks (height: number, callback: (err: Error | null, rows: any[]) => void)
    {
        ...
    }
```

The code after work
```
    public putBlocks (data: any, onSuccess: () => void, onError: (err: Error) => void)
    {
        ...
    }

    public getBlocks (height: number, onSuccess: (rows: any[]) => void, onError: (err: Error) => void)
    {
        ....
    }
```

Related to Issue #29 

There is no need for `err` to be `null` anymore.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -41,34 +41,36 @@ class Stoa {
                 return;
             }
 
-            this.ledger_storage.getValidatorsAPI(height, null, (err: Error | null, rows: any[]) =>
-            {
-                if (err != null)
+            this.ledger_storage.getValidatorsAPI(height, null,
+                (rows: any[]) =>
+                {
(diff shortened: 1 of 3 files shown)


TrustHenry:
I think we are currently combining commit with this PR. Is that right?

MichaelKim20:
> I think we are currently combining commit with this PR. Is that right?

It's better to separate it because the method and effect of implementing it are different.

TrustHenry (approved):
Nice! LGTM

TrustHenry:
This needs a rebase

Geod24 (approved):

PR #32, opened 2020-07-01
Fix the return as an object, not a list, in the /validator

TrustHenry:


src/Stoa.ts
```diff
@@ -98,18 +98,12 @@ class Stoa {
...
-                    for (const row of rows)
-                    {
-                        let preimage: IPreimage = {distance: row.distance, hash: '0'} as IPreimage;
-                        var validator: ValidatorData | null =
-                            new ValidatorData(row.address, row.enrolled_at, row.stake, preimage);
-                        out_put.push(validator);
-                    }
-                    res.status(200).send(JSON.stringify(out_put));
+                    let preimage: IPreimage = {distance: rows[0].distance, hash: '0'} as IPreimage;
+                    var validator: ValidatorData | null =
```
MichaelKim20:
```suggestion
                    let validator: ValidatorData | null =
```

MichaelKim20 (changes requested):
We'd better use 'let' rather than 'var' for variable declarations in the future. 
This is a reference to this.
English
https://www.typescriptlang.org/docs/handbook/variable-declarations.html
Korean
https://infoscis.github.io/2017/05/14/TypeScript-handbook-variable-declarations/

src/Stoa.ts
TrustHenry: That's a good point.

TrustHenry: I apply the suggestion now.

MichaelKim20 (approved):
LGTM

src/Stoa.ts
```diff
@@ -98,18 +98,12 @@ class Stoa {
                     return;
                 }
 
-                if (rows.length)
+                if (rows.length == 1)
```
Geod24, line 101: That's a weird change. What happen if we were to get `rows.length == 2` ?

src/Stoa.ts
```diff
@@ -98,18 +98,12 @@ class Stoa {
...
-
-                    for (const row of rows)
-                    {
-                        let preimage: IPreimage = {distance: row.distance, hash: '0'} as IPreimage;
-                        var validator: ValidatorData | null =
-                            new ValidatorData(row.address, row.enrolled_at, row.stake, preimage);
-                        out_put.push(validator);
-                    }
-                    res.status(200).send(JSON.stringify(out_put));
+                    let preimage: IPreimage = {distance: rows[0].distance, hash: '0'} as IPreimage;
```
Geod24, line 103: Ah that explains it :D

MichaelKim20:
I called the address below to get the following results.

```
http://localhost:3836/validator/GBUVRIIBMHKC4PE6BK7MO2O26U2NJLW4WGGWKLAVLAA2DLFZTBHHKOEK
```

```
{"address":"GBUVRIIBMHKC4PE6BK7MO2O26U2NJLW4WGGWKLAVLAA2DLFZTBHHKOEK","enrolled_at":0,"stake":"0x86f1a6dff3b1f2256d2417b71ecc5511293b224894da5fd75c192965aa1874824ca777ecac678c871e717ad38c295046f4f64130f31750aa967c30c35529944a","preimage":{"distance":0,"hash":"0"}}
```

Geod24:
What about the previous comment ?

src/Stoa.ts
TrustHenry: This is the function of returning a single validator value.
If `rows.length ==2` return 204 to HTTP response code.

src/Stoa.ts
TrustHenry: I will open another PR related to the preimage.

src/Stoa.ts
Geod24: But we look it up by address. We should NEVER have `res.length > 1`.

Geod24: Actually scratch that. Currently, it is possible that a validator uses multiple UTXOs to enroll.

TrustHenry: Oh, that's right.
I think I overlooked it.
I'll open a new PR.

TrustHenry: If so, does the validator of the function return an enrollment of one of the UTXOs?

Geod24: I would say all of them, potentially sorted.

Issue #37, opened 2020-07-13
Simple change is made using `Statement` of sqlite3


MichaelKim20:
When inserting multiple records into a database, the existing complex circular statement is simply changed using `Statement`.

ex)
``` sql
  var stmt = db.prepare("INSERT INTO users VALUES (?,?)");
  for (var i = 0; i < 10; i++) {
      stmt.run("user " + i, "email " + i);
  }
  stmt.finalize();
```

PR #43, opened 2020-07-20
Use a sqlite3's `Statement` to insert multiple records into the database

MichaelKim20:
I organized the code more simply using sqlite3's `Statement`.

I use three commits to solve this.
When used as one commit, Git could not recognize the change as a system.
I thought it would be difficult for reviewers to understand the change.
Maybe it's because the changes are too different.
Therefore, I had to separate Commit into three.
First Commit - Add New Method 1
Second Commit - Add New Method 2
Third Commit - Delete unused methods

Related to issue #37


TrustHenry (approved):
LGTM

Issue #39, opened 2020-07-20
Correction according to Agora block_externalized event


MichaelKim20:
Agora's issue https://github.com/bpfkorea/agora/issues/1033 uses the RestInterfaceClient in vibe.d.
It is necessary to analyze the data types of block data that this conveys and modify the existing code so that it works without any problems.

PR #54, opened 2020-07-24
JSON type block data is also acceptable and change API path for test

MichaelKim20:
Correction according to the RestInterfaceClient in vibe.d.

Related to issue #39


TrustHenry (approved):
Test error correction from the previous commit included.

Issue #60, opened 2020-07-30
When storing block data in a database, it is necessary to save storage space.


MichaelKim20:
Change Hash, Signature data from string to BLOB.

MichaelKim20:
I separated this issue into several new issues.

PR #61, opened 2020-07-30
Redefines the type of data as same as the block structure of the agora

MichaelKim20:
Hash, Signature, PublicKey changed to have binary data inside

Related to #60


MichaelKim20:
I implemented serialization and deserialization functions. I added it here with the test code.

MichaelKim20:
This is the code that extracts the genesis block and its serialized data from Agora's test code.
https://gist.github.com/MichaelKim20/1da4b94499e18e6c50f1336fe5321990

MichaelKim20:
I separated it into #93, #94, #95, #96, #97, considering the convenience of the review.
So I'll close it.

PR #62, opened 2020-07-31
Add the ability to request and receive blocks to the Agora node

### LINKED ISSUE #45: Implement the ability to request and recover missing blocks from Agora

MichaelKim20:
This will only be used for restoration purposes.

Libraries 'axios' and 'urijs' are changed from development
to production for data requests to Agora nodes

Related to #45

### DIFF
--- a/src/modules/agora/AgoraClient.ts
+++ b/src/modules/agora/AgoraClient.ts
@@ -1,0 +1,132 @@
+/*******************************************************************************
+
+    The class that recovers data
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import axios, { AxiosInstance, AxiosResponse, AxiosError } from 'axios';
+import URI from "urijs";
+
+/**
+ * The class that recovers data
+ */
+export class AgoraClient
+{
+    /**
+     * The network address to connect to Agora
+     */
+    private host: string;
+
+    /**
+     * The network port to connect to Agora
+     */
+    private port: string;
+
+    /**
+     * The instance of the axios
+     */
+    private client: AxiosInstance;
+
+    /**
+     * Constructor
+     * @param host - The network address to connect to Agora
+     * @param port - The network port to connect to Agora
+     */
+    constructor (host: string, port: string)
+    {
+        this.host = host;
+        this.port = port;
+        this.client = axios.create();
+        this.client.defaults.timeout = 10000;
+    }
+
+    /**
+     * Requests and receives data needed for recovery from Agora.
+     * @param block_height - The height of the block
+     * @param max_blocks - The maximum number of block to request
+     */
+    public requestBlocks (block_height: number, max_blocks: number): Promise<Array<any>>
+    {
+        return new Promise<Array<any>>((resolve, reject) =>
+        {
+            let uri = URI(this.host)
+                .port(this.port)
+                .directory("blocks_from")
+                .addSearch("block_height", block_height)
+                .addSearch("max_blocks", max_blocks);
+
+            this.client.get(uri.toString())
+                .then((response: AxiosResponse) =>
+                {
+                    if (response.status == 200)
+                    {
+                        resolve(response.data);
+                    }
+                    else
+                    {
+                        reject(new Error(response.statusText));
+                    }
+                })
+                .catch((reason: any) => {
+                    reject(handleNetworkError(reason));
+                });
+        });
+    }
+}
+
+/**
+ * Check if parameter `reason` is type `AxiosError`.
+ * @param reason{any} This is why the error occurred
+ * @returns {boolean}
+ */
+function isAxiosError (reason: any): reason is AxiosError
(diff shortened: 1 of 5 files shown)


src/modules/agora/AgoraClient.ts
```diff
@@ -0,0 +1,131 @@
...
+    private client: AxiosInstance;
+
+    /**
+     * Constructor
+     * @param host - The network address to connect to the Agora
+     * @param port - The network port to connect to the Agora
+     */
+    constructor (host: string, port: string)
+    {
+        this.client = axios.create();
```
TrustHenry, line 46:
```
this.client = axios.create({
            baseURL: `url:port`
        });
```

Don't it need baseURL basic settings for `Axios.create`?

TrustHenry: `this.client.defaults.timeout = 3000;`

It can also set a timeout.

MichaelKim20: Thank you for your good suggestion.
I changed it as below.

```
    /**
     * The URL encoder
     */
    private uri: URI;

    /**
     * Constructor
     * @param host - The network address to connect to the Agora
     * @param port - The network port to connect to the Agora
     */
    constructor (host: string, port: string)
    {
        this.client = axios.create();
        this.client.defaults.timeout = 10000;
        this.uri = URI(host).port(port);
    }
```

TrustHenry (approved):
LGTM

src/modules/agora/AgoraClient.ts
```diff
@@ -0,0 +1,125 @@
+/*******************************************************************************
+
+    The class that recover data
```
Geod24:
```suggestion
    The class that recovers data
```

src/modules/agora/AgoraClient.ts
```diff
@@ -0,0 +1,125 @@
...
+    private client: AxiosInstance;
+
+    /**
+     * The URL encoder
+     */
+    private uri: URI;
+
+    /**
+     * Constructor
+     * @param host - The network address to connect to the Agora
```
Geod24: You can either write "connect to Agora" or "connect to the Agora node", but not "connect to the Agora".
Likewise, you can say "going to Samsung" or "going to the Samsung headquarter", not "going to the Samsung".

src/modules/agora/AgoraClient.ts
```diff
@@ -0,0 +1,125 @@
...
+     * @param max_blocks - The maximum number of block to request
+     */
+    public requestBlocks (block_height: number, max_blocks: number): Promise<Array<any>>
+    {
+        return new Promise<Array<any>>((resolve, reject) =>
+        {
+            this.uri
+                .directory("blocks_from")
+                .setSearch("block_height", block_height.toString())
+                .setSearch("max_blocks", max_blocks.toString());
```
Geod24: If I understand correctly, this modifies the `uri` member ? Doesn't that mean that this function is not re-entrant, which is at odds with its `Promise`-based approach ?

Geod24 (changes requested):
Questions about the URI approach

src/modules/agora/AgoraClient.ts
MichaelKim20: The address and port of the client need not be preset. Because I have to enter the entire URL every time I call this.client.get.
So as an alternative, I set the address and port on the URL.

I checked in advance through the test.
I have confirmed that it is reset.
```
describe ('Test of Stoa API Server', () => {
    let host: string = 'http://localhost';
    let port: string = '3837';

    it('Test of URI', () => {
        let uri = URI(host)
            .port(port);

        uri
            .directory("1")
            .setSearch("block_height", "1")
            .setSearch("max_blocks", "1");
        assert.strictEqual(uri.toString(), "http://localhost:3837/1/?block_height=1&max_blocks=1");

        uri
            .directory("2")
            .setSearch("block_height", "2")
            .setSearch("max_blocks", "2");
        assert.strictEqual(uri.toString(), "http://localhost:3837/2/?block_height=2&max_blocks=2");
    });
});
```

src/modules/agora/AgoraClient.ts
MichaelKim20: I changed it

src/modules/agora/AgoraClient.ts
MichaelKim20: I changed it

src/modules/agora/AgoraClient.ts
Geod24: It's still here ?

MichaelKim20: Yeah, I revised the 18th line.

src/modules/agora/AgoraClient.ts
MichaelKim20: @Geod24 I'll check and ask for a review because there's something I need to revise.

MichaelKim20: @Geod24 That's right. The value of this.uri can be changed by other repeated calls.
So I changed it back to a local variable
Ready of review

Geod24 (approved):

PR #63, opened 2020-07-31
Add an error handler in `putBlocks`.

MichaelKim20:
When an error occurred, an error callback had to be executed, but it was not called.


TrustHenry (approved):
Nice catch!

Geod24:
This probably needs a test!

MichaelKim20:
> This probably needs a test!

That's right. I think I need it.
I'm going to try using Promise for the part that records in the database.
Then I'll think about how to do it then.

PR #65, opened 2020-07-31
Implement the function that provides the height of the block to be added next

### LINKED ISSUE #45: Implement the ability to request and recover missing blocks from Agora

MichaelKim20:
Stoa records data pushed from the Agora node in the database.
This is necessary to ensure that a block with a block height of 5 is correct before it is recorded.
If the block height is recorded up to 3 and just when the block height of 5 was pushed in,
This allows Stoa to detect that the expected 4 and 5 are different.

Related at #45

### DIFF
--- a/docs/Database_Schema.md
+++ b/docs/Database_Schema.md
@@ -171,3 +171,29 @@ CREATE TABLE "validators" (
     PRIMARY KEY("enrolled_at","utxo_key")
 )
 ```
+
+----
+
+## 6. Table **information**
+
+It can store information that is required for operation.
+The following data is recorded when the most recently recorded block height is 100.
+
+{"key": "height", "value": "100"}
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+|  key              | TEXT      | Y  | Y        |          | The key   |
+|  value            | TEXT      |    | Y        |          | The value |
+
+### _Create Script_
+
+```sql
+CREATE TABLE informations (
+    key         TEXT NOT NULL,
+    value       TEXT,
+    PRIMARY KEY(key)
+)
+```
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -13,8 +13,10 @@
 
 import { Hash }  from '../common/Hash'
 import { Storages } from './Storages';
-import { Block, Enrollment, Transaction,
-    TxInputs, TxOutputs } from '../data';
+import {
+    Block, Enrollment, Height, Transaction,
+    TxInputs, TxOutputs
+} from '../data';
 
 /**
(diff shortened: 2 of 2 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -518,4 +518,31 @@ export class LedgerStorage extends Storages
...
     }
+
+    /**
+     * Returns the height of the block to be added next
+     */
+    public getExpectedBlockHeight(): Promise<number>
+    {
+        return new Promise<number>((resolve, reject) =>
+        {
+            let sql = `select height from blocks order by height desc limit 1;`;
```
TrustHenry: For readability, Looks good to use DML of SQL in capital letters.

MichaelKim20: I changed it as below.
`SELECT height FROM blocks ORDER BY height DESC LIMIT 1;`

AndrejMitrovic:
Isn't it better to just store the latest block height somewhere in the database? Doing an `order by` every single time a block is received is expensive, is it not?

MichaelKim20:
> Isn't it better to just store the latest block height somewhere in the database? Doing an `order by` every single time a block is received is expensive, is it not?

That's what I thought at first, but I tried to implement the function with a small change.
By the way, I think performance is more important, so I'll revise it according to your opinion.

MichaelKim20:
I added a table called information at @AndrejMitrovic 's suggestion and had the last block height stored there.

src/modules/storage/LedgerStorage.ts
```diff
@@ -518,4 +533,51 @@ export class LedgerStorage extends Storages
 
         this.query(sql, [], onSuccess, onError);
     }
+
+    /**
+     * Puts a block to database
```
TrustHenry: It would be better to explain the height of the block successfully stored.

docs/Database_Schema.md
```diff
@@ -171,3 +171,24 @@ CREATE TABLE "validators" (
...
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+|  key              | TEXT      | Y  | Y        |          | The key   |
+|  value            | TEXT      |    | Y        |          | The value |
+
+### _Create Script_
+
+```sql
+CREATE TABLE informations (
```
TrustHenry, line 194: The Table name alone is difficult to estimate the purpose of this table.
It would be better to add a more specific table name or description.

MichaelKim20: I changed it

src/modules/storage/LedgerStorage.ts
MichaelKim20: I changed it

TrustHenry (approved):
LGTM
I can merge by after rebase
This blocks #68 PR.

MichaelKim20:
Rebased

Issue #44, opened 2020-07-20
Ensure that the same block data is processed only once.


MichaelKim20:
It can be pushed several times for the same block in the Agora.
Therefore, we need to make sure that block data that has already been processed is not processed.
Whether or not the same block is simply determined by the height of the block.

PR #68, opened 2020-08-03
Recovery and testing of missing blocks on receipt of blocks

### LINKED ISSUE #45: Implement the ability to request and recover missing blocks from Agora

MichaelKim20:
It uses code made in PR #62 and PR #65
These will be used after they are merged first.
The test will be carried out in the following steps

Related #45, #44

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -6,20 +6,54 @@ import { ValidatorData, IPreimage, IValidator } from "./modules/data/ValidatorDa
 import { Hash } from "./modules/common/Hash";
 import { PreImageInfo } from "./modules/data";
 import { cors_options } from "./cors";
+import { AgoraClient } from "./modules/agora/AgoraClient";
+import { TaskManager } from "./modules/task/TaskManager";
 
 class Stoa {
     public stoa: express.Application;
 
     public ledger_storage: LedgerStorage;
 
-    constructor (file_name: string) {
+    /**
+     * The network host to connect to Agora
+     */
+    private readonly agora_host: string;
+
+    /**
+     * The network port to connect to Agora
+     */
+    private readonly agora_port: string;
+
+    /**
+     * The temporary storage of the data received from Agora
+     */
+    protected pool: Array<IPooledData>;
+
+    /**
+     * The task manager that periodically executes registered a task function
+     */
+    protected task_manager: TaskManager;
+
+    /**
+     * Constructor
+     * @param database_filename sqlite3 database file name
+     * @param agora_host The network host to connect to Agora
+     * @param agora_port The network port to connect to Agora
+     */
+    constructor (database_filename: string, agora_host: string, agora_port: string)
+    {
+        this.agora_host = agora_host;
+        this.agora_port = agora_port;
+
+        this.pool = [];
+
         this.stoa = express();
         // parse application/x-www-form-urlencoded
-        this.stoa.use(bodyParser.urlencoded({ extended: false }))
+        this.stoa.use(bodyParser.urlencoded({ extended: false }));
         // parse application/json
         this.stoa.use(bodyParser.json());
         // create blockStorage
-        this.ledger_storage = new LedgerStorage(file_name, (err: Error | null) =>
+        this.ledger_storage = new LedgerStorage(database_filename, (err: Error | null) =>
         {
             if (err != null)
             {
@@ -31,6 +65,12 @@ class Stoa {
         this.stoa.use(cors(cors_options));
         // enable pre-flight
         this.stoa.options('*', cors(cors_options));
+
+        // create task manager, delay time is 10 milliseconds
+        this.task_manager = new TaskManager(() =>
+        {
+            return this.task();
+        }, 10);
 
         /**
          * Called when a request is received through the `/validators` handler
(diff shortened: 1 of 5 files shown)


src/Stoa.ts
```diff
@@ -147,22 +156,102 @@ class Stoa {
...
-                () =>
+                // Obtain the height of the block that should be recorded in the database.
+                let expected_height = await this.ledger_storage.getExpectedBlockHeight();
+                if (height == expected_height)
                 {
+                    // The Normal case
+
+                    // Save the block just received
+                    await save(this.ledger_storage, block);
+                    this.pushSaveLog(`save ${height}`);
```
TrustHenry: `save 1`
How would you describe it
`Block storage success #1`

MichaelKim20: This will only be used for verification in the test code.
Only the declaration of pushSaveLog, no implementation.
Its implementation is only in the test code.

TrustHenry:
```suggestion
                    this.pushSaveLog(`Block storage success # ${height}`);
```
Yes, we'd better make it into a more specific expression of the log.

MichaelKim20: I solved this IRL.

MichaelKim20: I just added a log output function separately.

MichaelKim20:
I improved so that when a new block arrives while the block is being recorded, the ongoing work is not interrupted.
A. Blocks of block heights 0 and 1 are received.
B. Block 3 is received. Block 2 will be restored at this time. Block 3 is recorded.
C. Block 4 is received.  Block 4 is recorded.
D. Block 3 is received.  Block 3 is ignored.
**When Block 4 arrives while Task B is in progress**

src/Stoa.ts
```diff
@@ -2,16 +2,25 @@ import express from "express";
...
 class Stoa {
     public stoa: express.Application;
 
     public ledger_storage: LedgerStorage;
 
-    constructor (file_name: string) {
+    private readonly agora_address: string;
+    private readonly agora_port: string;
+
+    constructor (file_name: string, agora_address: string, agora_port: string)
```
Geod24: `agora_address` => `agora_host`
It would also be nice to rename `file_name` to something that says what it is, and is more accurate (it is a path, not a file name, and you can have multiple parameters which are path, but this one is used for the ledger storage)

src/Stoa.ts
```diff
@@ -147,18 +156,97 @@ class Stoa {
...
+                    (block.header.height !== undefined) &&
+                    (block.header.height.value !== undefined)
+                )
+                {
+                    return Number(block.header.height.value);
+                }
+                else
+                {
+                    throw Error("Not found block height in JSON Block");
+                }
```
Geod24: A style note: It is better to get the error cases "out of the way". It also allows to remove nesting.
In this case, since `if` will `return`, you can just write:
```suggestion
                if  (
                    (block.header !== undefined) &&
                    (block.header.height !== undefined) &&
                    (block.header.height.value !== undefined)
                )
                {
                    return Number(block.header.height.value);
                }
                throw Error("Not found block height in JSON Block");
```

But then, it's a bit odd, hence my recommendation to handle errors first:
```suggestion
                if  (error_condition)
                    throw Error("Not found block height in JSON Block");
                return Number(block.header.height.value);
```

src/Stoa.ts
```diff
@@ -147,18 +156,97 @@ class Stoa {
...
-                console.error("Failed to store the payload of a push to the DB: " + err);
+            try
+            {
+                let height = getBlockHeight(block);
+
+                // Obtain the height of the block that should be recorded in the database.
+                let expected_height = await this.ledger_storage.getExpectedBlockHeight();
+                if (height == expected_height)
+                {
+                    // The Normal case
```
Geod24: Why the uppercase on normal ?

src/Stoa.ts
```diff
@@ -147,18 +156,97 @@ class Stoa {
...
+                    // Save a block just received
+                    await this.ledger_storage.putBlocks(block);
+                    this.pushTestEvent(`save ${height}`);
```
Geod24: This is rather specialized, and IMO something we should avoid. Having for-test-only methods usually leads to issues because you have different code path for tests and "real" code.

src/Stoa.ts
```diff
@@ -147,18 +168,124 @@ class Stoa {
...
+                    // count is `block_data.header.height - expected_height`
+                    // Record a current block after recovery.
+
+                    this.saving_block = true;
+
+                    let block_height = expected_height;
+                    let max_blocks = height - expected_height;
+
+                    let agora_client = new AgoraClient(this.agora_address, this.agora_port);
+                    let blocks = await agora_client.requestBlocks(block_height, max_blocks);
```
Geod24: The recovery code should not be in the handler. This could lead to infinite recursion very easily.

src/Stoa.ts
```diff
@@ -147,18 +168,124 @@ class Stoa {
...
+            {
+                this.saving_block = false;
+            }
         });
     }
+
+    protected pushTestEvent (message: string)
+    {
+
+    }
```
Geod24: A method that does nothing, and is only there for the sake of the derived class, which is there for testing, should be a red flag. This is different from dependency injection because the base method does something.

Geod24 (changes requested):
Two main issues:
1) The testing approach is not good;
2) The recovery shouldn't be done in a blocking manner from the handler. It can be *scheduled*, but it shouldn't block on it.

MichaelKim20:
When there was already work in progress, I added a pool for smoother processing.
@Geod24 Ping

MichaelKim20:
@Geod24 Ready for review

Geod24:
Ping me when this is ready.

MichaelKim20:
@Geod24 I rebased it. ready for review.

src/Stoa.ts
```diff
@@ -228,34 +254,186 @@ class Stoa {
...
-            .catch((err) => {
-                console.error("Failed to store the payload of a update to the DB: " + err);
-                res.status(500).send("An error occurred while update");
-            });
+                    await this.ledger_storage.updatePreImage(pre_image);
+                    console.log(`Saved a pre-image enroll_key : ${pre_image.enroll_key.substr(0, 18)}, ` +
+                        `hash : ${pre_image.hash.substr(0, 18)}, distance : ${pre_image.distance}`);
+                    resolve();
+                }
+                catch(err)
```
Geod24, line 420:
```suggestion
                catch (err)
```

src/main.ts
```diff
@@ -3,7 +3,7 @@ import Stoa from './Stoa';
 import express from "express";
 
 const port: number = Number(process.env.PORT) || 3836;
-const stoa: express.Application = new Stoa("database").stoa;
+const stoa: express.Application = new Stoa("database", "127.0.0.1", "2826").stoa;
```
Geod24, line 6: We should make this configurable in a future PR.

Geod24 (approved):
LGTM

Issue #51, opened 2020-07-22
Apply DB transactions to the Block storage processing


TrustHenry:
If an error occurs while storing blocks and transactions or enrollments, that does not match the ledger actual data may exist.
For example, if only block data exists.
In this case, rollback must be done and must be restored in the tables.

PR #70, opened 2020-08-04
Apply DB transactions to the Block storage processing

TrustHenry:
If an error occurs while storing blocks and transactions or enrollments, that do not match the ledger actual data may exist.
For example, if only block data exists.
In this case, rollback must be done and must be `rollback` in the tables.

Related #51

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -153,12 +153,15 @@ export class LedgerStorage extends Storages
                 {
                     let block: Block = new Block();
                     block.parseJSON(data);
+                    await this.begin();
                     await saveBlock(this, block);
                     await this.putTransactions(block);
                     await this.putEnrollments(block);
+                    await this.commit();
                 }
                 catch (error)
                 {
+                    await this.rollback();
                     reject(error);
                     return;
                 }
--- a/src/modules/storage/Storages.ts
+++ b/src/modules/storage/Storages.ts
@@ -62,4 +62,62 @@ export class Storages
     {
         this.db.close();
     }
+
+    /**
+     * SQLite transaction statement
+     * To start a transaction explicitly,
+     * Open a transaction by issuing the begin function
+     * the transaction is open until it is explicitly
+     * committed or rolled back.
+     */
+    protected begin (): Promise<void>
+    {
+        return new Promise<void>((resolve, reject) =>
+        {
+            this.db.run('BEGIN', (err: Error | null) =>
+            {
+                if (err == null)
+                    resolve();
+                else
+                    reject(err);
+            });
+        });
+    }
+    
+    /**
+     * SQLite transaction statement
+     * Commit the changes to the database by using this.
+     */
+    protected commit (): Promise<void>
+    {
+        return new Promise<void>((resolve, reject) =>
+        {
+            this.db.run('COMMIT', (err: Error | null) =>
+            {
+                if (err == null)
+                    resolve();
+                else
+                    reject(err);
+            });
+        });
+    }
+
+    /**
+     * SQLite transaction statement
+     * If it do not want to save the changes,
+     * it can roll back using this.
+     */
+    protected rollback (): Promise<void>
+    {
+        return new Promise<void>((resolve, reject) =>
+        {
+            this.db.run('ROLLBACK', (err: Error | null) =>
+            {
+                if (err == null)
+                    resolve();
+                else
+                    reject(err);
+            });
+        });
+    }
 }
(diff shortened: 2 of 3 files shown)


TrustHenry:
This require #71. Rebase operation is required after the merge.

TrustHenry:
Ready for review.

src/modules/storage/Storages.ts
```diff
@@ -62,4 +62,60 @@ export class Storages
...
+            this.db.run('BEGIN', (err: Error | null) =>
+            {
+                if (err == null)
+                    resolve();
+                else
+                    reject(err);
+            });
+        });
+    }
+    /**
```
MichaelKim20, line 86:
```suggestion
    }
    
    /**
```

src/modules/storage/Storages.ts
```diff
@@ -62,4 +62,60 @@ export class Storages
...
+            this.db.run('COMMIT', (err: Error | null) =>
+            {
+                if (err == null)
+                    resolve();
+                else
+                    reject(err);
+            });
+        });
+    }
+    /**
```
MichaelKim20:
```suggestion
    }
    
    /**
```

src/modules/storage/Storages.ts
MichaelKim20:
```suggestion
    }
    
    /**
```

tests/Storage.test.ts
```diff
@@ -293,4 +293,33 @@ describe('LedgerStorage', () =>
...
+    {
+        let ledger_storage = new LedgerStorage(":memory:", async (err1: Error | null) =>
+        {
+            assert.ok(!err1, err1?.message);
+
+            let block = new Block();
+            block.parseJSON(sample_data[0]);
+
+            await ledger_storage.putEnrollments(block);
+            assert.rejects(ledger_storage.putBlocks(sample_data[0]),
```
MichaelKim20:
```suggestion
            await assert.rejects(ledger_storage.putBlocks(sample_data[0]),
```
There's another test left. `await` is required.

tests/Storage.test.ts
```diff
@@ -293,4 +293,33 @@ describe('LedgerStorage', () =>
...
+            // Make sure that the block is not stored because it has been rolled back.
+            ledger_storage.getBlocks(0,
+                (rows: any[]) =>
+                {
+                    assert.strictEqual(rows.length, 0);
+                },
+                (err: Error) =>
+                {
+                    assert.ok(!err, err?.message);
+                });
```
MichaelKim20: If this is done, we can get a clearer test result.
And we need to check if it works properly after rollback.
```suggestion
            let getBlock = ((height: number): Promise<any[]> =>
            {
                return new Promise<any[]>((resolve, reject) =>
                {
                    ledger_storage.getBlocks(height,
                        (rows: any[]) => {
                            resolve(rows);
                        },
                        (err: Error) => {
                            reject(err);
                        });
                });
            });
            let rows0 = await getBlock(0);
            assert.strictEqual(rows0.length, 0);
           
            await ledger_storage.putTransactions(block);
            let getTransaction = ((height: number): Promise<any[]> =>
            {
                return new Promise<any[]>((resolve, reject) =>
                {
                    ledger_storage.getTransactions(height,
                        (rows: any[]) => {
                            resolve(rows);
                        },
                        (err: Error) => {
                            reject(err);
                        });
                });
            });
            let rows1 = await getTransaction(0);
            assert.strictEqual(rows1.length, 4);
```

MichaelKim20 (commented):
It's very good.
I'd like to add some more to the test code.

tests/Storage.test.ts
TrustHenry: It's a clearer test case.
I'll apply it as suggested.

TrustHenry:
Thank you.
I applied @MichaelKim20's  suggestion.

src/modules/storage/LedgerStorage.ts
```diff
@@ -153,12 +153,15 @@ export class LedgerStorage extends Storages
                 {
                     let block: Block = new Block();
                     block.parseJSON(data);
+                    await this.begin();
                     await saveBlock(this, block);
                     await this.putTransactions(block);
                     await this.putEnrollments(block);
+                    await this.commit();
```
Geod24, line 160: It seems very backward to have an await per line here. Those operations are dependent on one another, so they should be in the same promise.

Geod24 (approved):

src/modules/storage/LedgerStorage.ts
TrustHenry: Yes, that's right.
I think it can do a common, dependent task within Begin.

PR #73, opened 2020-08-10
Only create information table if it does not already exist

### LINKED ISSUE #72: Server fails to start correctly due to sql table creation error
If the **information** table already exists then the following error is reported when running the app via 
`npm start`
[Error: SQLITE_ERROR: table information already exists] {
  errno: 1,
  code: 'SQLITE_ERROR'
}

hewison-chris:
If information table is already existing then the stoa app reports error on startup.


MichaelKim20 (approved):
Thank you for correcting the bug.

PR #74, opened 2020-08-11
Stoa supports CORS and define CORS policies

### LINKED ISSUE #46: Stoa has been blocked by CORS policy
When accessing Stoa through a browser, it is not possible to request normally due to the CORS policy.
Solve this

TrustHenry:
This supports the preflight of the CORS-enabled browser or server.

Access-Control-Allow-Origin
cors policy for public API : All Allow

Related to #74


TrustHenry:
Rebase done.

src/cors.ts
```diff
@@ -0,0 +1,31 @@
...
+import cors from "cors";
+
+// cors policy for agora node
+export const internal_cors_options: cors.CorsOptions = {
+    allowedHeaders: '*',
+    credentials: true,
+    methods: "POST",
+    origin: 'http://127.0.0.1:2826',
+    preflightContinue: false
+};
```
MichaelKim20: It's hard to understand why this is necessary.

src/Stoa.ts
```diff
@@ -127,7 +129,7 @@ class Stoa {
          * When a request is received through the `/push` handler
          * JSON block data is parsed and stored on each storage.
          */
-        this.stoa.post("/block_externalized",
+        this.stoa.post("/block_externalized", cors(internal_cors_options),
```
MichaelKim20: What is the reason for using this?

MichaelKim20 (commented):
I have a question that I don't understand.

src/cors.ts
TrustHenry: I don't think it is necessary to expose POST /block_externalized to the outside, so I set a different policy than /GET.

MichaelKim20: I think this will not solve the problem you mentioned.
Also, I think it is different from the current issue.
How about solving other issues?

TrustHenry: Yes.  Another issue related to this has to be created.  For example, open another port and divid stoa

TrustHenry: I created an issue related to this. #76 
Current CORS policies can be taken in common.

TrustHenry:
The CORS policy was applied in common with all Origin Allowance of GET, POST.

src/Stoa.ts
TrustHenry: ^

MichaelKim20 (approved):
Good work

PR #77, opened 2020-08-13
Apply Promise to methods in LedgerStorage and test code

### LINKED ISSUE #69: Improve by using Promise for methods of LedgerStorage
Using async and await, we can easily write repeating sentences without using the nested callback.
But to take advantage of this, we must define the `Promise<return type>` as a return function.
But the methods of LedgerStorage are not. 
These are modified as functions that return `Promise<return type>`.
This will provide various choices in the use of these methods.

Example of use of `async / await / Promis`

``` TypeScript
async function putAllBlockData (ledger_storage: LedgerStorage,
    callback: (err: Error | null) => void)
{
    // Use Promise to record block data synchronously.
    let save = (storage: LedgerStorage, block: any) =>
    {
        return new Promise<boolean>((resolve, reject) => {
            storage.putBlocks(block, ()=>{resolve(true);}, (error)=>{reject(error);});
        });
    };

    for (let elem of sample_data)
    {
        try
        {
            await save(ledger_storage, elem);
        }
        catch (error)
        {
            callback(error);
            return;
        }
    }
    callback(null);
}
```

MichaelKim20:
I applied Promise not to use the nested callback.
I applied it to the method of LedgerStorage first.
And I applied this to the test code.

Linked #69

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -88,8 +88,8 @@ class Stoa {
                 return;
             }
 
-            this.ledger_storage.getValidatorsAPI(height, null,
-                (rows: any[]) =>
+            this.ledger_storage.getValidatorsAPI(height, null)
+                .then((rows: any[]) =>
                 {
                     if (rows.length)
                     {
@@ -136,12 +136,11 @@ class Stoa {
                     {
                         res.status(204).send();
                     }
-                },
-                (err: Error) =>
+                })
+                .catch((err) =>
                 {
                     console.error("Failed to data lookup to the DB: " + err);
                     res.status(500).send("Failed to data lookup");
-                    return;
                 }
             );
         });
@@ -165,8 +164,8 @@ class Stoa {
                 return;
             }
 
-            this.ledger_storage.getValidatorsAPI(height, address,
-                (rows: any[]) =>
+            this.ledger_storage.getValidatorsAPI(height, address)
+                .then((rows: any[]) =>
                 {
                     if (rows.length)
                     {
@@ -213,12 +212,11 @@ class Stoa {
                     {
                         res.status(204).send();
                     }
-                },
-                (err: Error) =>
+                })
+                .catch((err) =>
                 {
                     console.error("Failed to data lookup to the DB: " + err);
                     res.status(500).send("Failed to data lookup");
-                    return;
                 }
             );
         });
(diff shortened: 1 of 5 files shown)


TrustHenry:
I think we should add a function description of the Promise return value.
To help users understand the get functions or functions,

src/Stoa.ts
```diff
@@ -22,7 +22,6 @@ class Stoa {
             if (err != null)
             {
                 console.error(err);
-                throw new Error(err.message);
```
Geod24: Why would the ctor not throw if there's an error ?

Geod24:
Love it !

TrustHenry:
There are test codes that have not been applied due to the new code merge.
Could you Apply `Promise` it together?
https://github.com/bpfkorea/stoa/blob/42970f554626279046b0bfc84e904d3c47fa8c7c/tests/Storage.test.ts#L359

src/modules/storage/LedgerStorage.ts
```diff
@@ -33,11 +33,8 @@ export class LedgerStorage extends Storages
 
     /**
      * Creates tables related to the ledger.
-     * @param callback If provided, this function will be called when
-     * the database was finished successfully or when an error occurred.
-     * The first argument is an error object.
```
Geod24, line 34: Need to document the return value now

src/modules/storage/LedgerStorage.ts
```diff
@@ -183,21 +179,16 @@ export class LedgerStorage extends Storages
     /**
      * Gets a block data
      * @param height the height of the block to get
-     * @param onSuccess This function will be called when
-     * the database was finished successfully
-     * @param onError This function will be called when
-     * an error occurred.
```
Geod24, line 186: Ditto documentation on return value

Geod24:
Missing a lot of doc upgrades. I  commented on 2 occurrences, but there are much more that need fixing.

src/Stoa.ts
MichaelKim20: It's my mistake. I added it again.

MichaelKim20:
@Geod24 I rebased it, and I added a document.

TrustHenry (approved):
LGTM

MichaelKim20:
> LGTM

Thank you for review

MichaelKim20:
I rebased it

PR #86, opened 2020-08-14
Separate hashing function from class `Hash`

### LINKED ISSUE #79: Separate hashing function from class `Hash`
The class simplifies having a binary data of hash results of the hash.

MichaelKim20:
I changed the class Hash to have only the hash value of the binary.

The binary of hash is big-endian.
When this is converted to a hex string, it should be treated as a little-endian.
Previously, the binary was reversed and the binary was used.
I made it processed by changing the order when converted to hex string.


src/modules/storage/LedgerStorage.ts
```diff
@@ -500,9 +496,9 @@ export class LedgerStorage extends Storages
 
                         for (let out_idx = 0; out_idx < block.txs[tx_idx].outputs.length; out_idx++)
                         {
-                            this.hash.makeUTXOKey(block.merkle_tree[tx_idx], BigInt(out_idx));
+                            let h = makeUTXOKey(Hash.createFromString(block.merkle_tree[tx_idx]), BigInt(out_idx));
```
```suggestion
                            let utxo_key = makeUTXOKey(Hash.createFromString(block.merkle_tree[tx_idx]), BigInt(out_idx));
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -500,9 +496,9 @@ export class LedgerStorage extends Storages
...
                             await save_output(this, block.header.height.value, tx_idx, out_idx,
-                                block.merkle_tree[tx_idx], this.hash.toHexString(), block.txs[tx_idx].outputs[out_idx]);
+                                block.merkle_tree[tx_idx], h.toString(), block.txs[tx_idx].outputs[out_idx]);
```
```suggestion
                                block.merkle_tree[tx_idx], utxo_key.toString(), block.txs[tx_idx].outputs[out_idx]);
```

MichaelKim20:
I rebased it

Geod24 (approved):
Like it!

PR #87, opened 2020-08-14
Fix the style and update a field option of table `information`

MichaelKim20:
Fixed the style of the table creation script.
Updated a field option of table `information`(Added 'NOT NULL').


TrustHenry (approved):
Pretty neat.

Issue #59, opened 2020-07-29
We need the ability to manage logs that occur during execution.


MichaelKim20:
**Necessity**
If an exception occurs during a process run, it should be possible to recover unsaved data leaving logs.

**Simple requirement**
It shall have the ability to specify the type of log.
It shall be possible to set the maximum size of the log file.
The output of the log can be set to a file and console.

**Library candidates**
Winston / bunyan  /  log4js / typescript-logging / morgan

PR #89, opened 2020-08-14
Add a new Logger Class

TrustHenry:
Logger class is based on 'winston'.

It operates as test, development, and production
according to the environment variables 'process.env.NODE_ENV' mode.

This can replace the existing console.log.
The application of the existing code is applied to other PR.

Relate to #59 part 1

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,10 @@
     "_test": "mocha -r ts-node/register tests/**/*.test.ts",
     "_build": "webpack --config webpack.config.js",
     "_start": "nodemon --watch src --delay 1 --exec 'ts-node' src/main.ts",
-    "test": "npm-run-all --serial json2ts _test",
+    "test": "NODE_ENV=test npm-run-all --serial json2ts _test",
+    "debug": "NODE_ENV=development npm-run-all --serial json2ts _test",
     "build": "npm-run-all --serial json2ts _build",
-    "start": "npm-run-all --serial json2ts _start",
+    "start": "NODE_ENV=production npm-run-all --serial json2ts _start",
     "tsc": "tsc"
   },
   "repository": {
@@ -25,10 +26,12 @@
   "homepage": "https://github.com/bpfkorea/stoa#readme",
   "dependencies": {
     "@ctrl/ts-base32": "^1.2.3",
+    "@types/app-root-path": "^1.2.4",
     "@types/cors": "^2.8.7",
     "@types/sodium-native": "^2.3.5",
     "@types/sqlite3": "^3.1.6",
     "ajv": "^6.12.3",
+    "app-root-path": "^3.0.0",
     "assert": "^2.0.0",
     "axios": "^0.19.2",
     "body-parser": "^1.19.0",
@@ -39,7 +42,8 @@
     "sodium-native": "^3.2.0",
     "source-map-support": "^0.5.19",
     "sqlite3": "^4.2.0",
-    "urijs": "^1.19.2"
+    "urijs": "^1.19.2",
+    "winston": "^3.3.3"
   },
   "devDependencies": {
     "@types/chai": "^4.2.12",
--- a/src/modules/common/Logger.ts
+++ b/src/modules/common/Logger.ts
@@ -1,0 +1,78 @@
+/*******************************************************************************
+
+    Logging class for Stoa
+
+    Define the output format of Log and support console output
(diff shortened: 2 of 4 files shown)


src/modules/common/Logger.ts
```diff
@@ -0,0 +1,78 @@
...
+      handleExceptions: true,
+      json: false,
+      colorize: false,
+      format: combine(
+        label({ label: 'Stoa' }),
+        timestamp(),
+        logFormat
+      )
+    }
+}
```
MichaelKim20:
```suggestion
};
```

src/modules/common/Logger.ts
```diff
@@ -0,0 +1,78 @@
...
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import winston from "winston";
+import { format, transports } from "winston";
```
MichaelKim20:
```suggestion
import { format } from "winston";
```
`transports` are not used.
 You can replace `winston.transports` with `transports`.

If you use `winston.format`, I don't think you need to import.

src/modules/common/Logger.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+const options = {
+    // write log file options
+    file: {
+      filename: `${appRoot}/logs/Stoa.log`, // TODO: move to config
+      handleExceptions: true,
+      json: false,
+      maxsize: 10485760, // 10MB
+      maxFiles: 10,
```
MichaelKim20, line 33: The filename doesn't change depending on the date, do I need this `maxFiles`?
What happens if the file size is more than 10 MByte?
If the file size is more than 10 MByte and the log is not recorded, it would be nice to have the file renamed every day.

package.json
```diff
@@ -8,9 +8,9 @@
     "_test": "mocha -r ts-node/register tests/**/*.test.ts",
     "_build": "webpack --config webpack.config.js",
     "_start": "nodemon --watch src --delay 1 --exec 'ts-node' src/main.ts",
-    "test": "npm-run-all --serial json2ts _test",
+    "test": "NODE_ENV=test npm-run-all --serial json2ts _test",
```
MichaelKim20, line 11: "debug": "NODE_ENV=development npm-run-all --serial json2ts _test",
How about adding one more?

MichaelKim20 (dismissed):
Okay. Please check a few things.
It is also necessary to add a log folder to the .gitignore

src/modules/common/Logger.ts
TrustHenry: > Okay. Please check a few things.
> It is also necessary to add a log folder to the .gitignore

https://github.com/bpfkorea/stoa/blob/f5aced1a1b93e91ccac6c290d933826e1a2a43d7/.gitignore#L1-L3

It's already applied.

src/modules/common/Logger.ts
TrustHenry: Okay!

TrustHenry:
> It is also necessary to add a log folder to the .gitignore

https://github.com/bpfkorea/stoa/blob/f5aced1a1b93e91ccac6c290d933826e1a2a43d7/.gitignore#L1-L3
It's already applied.

package.json
TrustHenry: That's a good idea.

src/modules/common/Logger.ts
TrustHenry: There is a good library related to this.
https://www.npmjs.com/package/winston-daily-rotate-file
Do you want to be managed like Web log?

src/modules/common/Logger.ts
TrustHenry: I think you have a good tool.

src/modules/common/Logger.ts
Geod24: I think it'll be better to setup something like `logrotate` so that is handled outside of the `stoa` node.

Geod24 (approved):
LGTM, nice!

Geod24:
Just needs a rebase

src/modules/common/Logger.ts
MichaelKim20: I think Stoa can run outside of Linux.

PR #91, opened 2020-08-20
Remove Unnecessary webpack

TrustHenry:
Webpack is not required because Stoa operates on serverside.


.gitignore
```diff
@@ -103,9 +103,6 @@ dist
 # TernJS port file
 .tern-port
 
-# webpack build ouputs
-.dist
```
MichaelKim20, line 107: I think we need to get rid of the `dist`.

TrustHenry: It's already applied to PR.
Does it mean anything else?

MichaelKim20 (approved):

PR #93, opened 2020-08-22
Redefines the type of data as same as the block structure of the agora

### LINKED ISSUE #81: Redefines the type of data as same as the block structure of the agora
Use an instance of the class Hash instead of a hash string.
Use an instance of the class Signature instead of a signature string.
Use an instance of the class PublicKey instead of a public key string.

MichaelKim20:
Use an instance of the class Hash instead of a hash string.
Use an instance of the class Signature instead of a signature string.
Use an instance of the class PublicKey instead of a public key string.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -430,8 +430,8 @@ class Stoa {
                     pre_image.parseJSON(stored_data.data);
 
                     await this.ledger_storage.updatePreImage(pre_image);
-                    logger.info(`Saved a pre-image enroll_key : ${pre_image.enroll_key.substr(0, 18)}, ` +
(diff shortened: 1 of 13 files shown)


src/modules/data/Hash.ts
```diff
@@ -42,8 +43,11 @@ export class Hash
     constructor (bin?: Buffer)
     {
         this.data = Buffer.alloc(Hash.Width);
-        if (bin != undefined)
+        if (bin !== undefined)
+        {
+            assert.strictEqual(bin.length, Hash.Width);
```
TrustHenry: How does It handle the error here?

MichaelKim20: The following error is thrown:
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

PR #94, opened 2020-08-22
Implement the function to obtain the hash value of block data

### LINKED ISSUE #83: Implement functions such as hashFull and hashPart in Agora
It is necessary to calculate the hash of various data.
It will also be used in various places.

MichaelKim20:
Implement the function to obtain the hash value of block data

Relates to #83

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3146,6 +3146,11 @@
         }
       }
     },
+    "smart-buffer": {
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/smart-buffer/-/smart-buffer-4.1.0.tgz",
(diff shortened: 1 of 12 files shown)


src/modules/data/BitField.ts
```diff
@@ -14,23 +14,40 @@
...
  * Convert JSON object to TypeScript's instance.
  * An exception occurs if the required property are not present.
  */
 export class BitField
 {
-    _storage: number[] = [];
+    /**
+     * The storage with bit data
+     */
+    public _storage: number[];
```
TrustHenry: Is there a reason to start with the under-bar in the public variable?
I can't understand the style convention.

MichaelKim20: This is using `_storage` to keep it the same as the JSON format received from Agora.

PR #97, opened 2020-08-22
Add the ability to store `Hash` and `Signature` in the database as binary

### LINKED ISSUE #60: When storing block data in a database, it is necessary to save storage space.

MichaelKim20:
I added the ability to store `Hash` and `Signature` in the database as binary

### DIFF
--- a/src/modules/data/BlockHeader.ts
+++ b/src/modules/data/BlockHeader.ts
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    The class that defines and parses the header of a block.
+    The class that defines the header of a block.
 
     Copyright:
         Copyright (c) 2020 BOS Platform Foundation Korea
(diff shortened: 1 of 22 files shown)


Geod24:
Hum, I would imagine that it's better to have it as text, as it's easier for human to inspect. But if you prefer binary and one just have to use stoa, I have no  strong opinion.

AndrejMitrovic:
> Hum, I would imagine that it's better to have it as text, as it's easier for human to inspect. But if you prefer binary and one just have to use stoa, I have no strong opinion.

That's why the commit message should explain **why** the change was made.

MichaelKim20:
> > Hum, I would imagine that it's better to have it as text, as it's easier for human to inspect. But if you prefer binary and one just have to use stoa, I have no strong opinion.
> 
> That's why the commit message should explain **why** the change was made.

I added the following to the commit message.
```
Its purpose is to save database storage space.
Use the hex function in the query to obtain a string of hexadecimal digits.
```

MichaelKim20:
@Geod24 @AndrejMitrovic Ready for review

src/modules/data/BlockHeader.ts
```diff
@@ -120,4 +121,17 @@ export class BlockHeader
 
         return this;
     }
+
+    /**
+     * Collects data to create a hash.
+     * @param buffer - The buffer where collected data is stored
+     */
+    public computeHash (buffer: SmartBuffer)
```
AndrejMitrovic, line 129: Why is this needed? We already have the hash in the header, why does stoa have to generate it again?

MichaelKim20:
@Geod24 If you don't think you need much of this PR function, please review the previous PR.

src/modules/utils/buffer.ts
```diff
@@ -15,38 +15,85 @@
...
+export function writeToString (source: Buffer, endian?: Endian): string
 {
-    let hex = [];
-    for (let idx = source.length-1; idx >= 0; idx--)
+    if ((endian === undefined) || (endian == Endian.Little))
     {
-        hex.push((source[idx] >>> 4).toString(16));
-        hex.push((source[idx] & 0xF).toString(16));
+        let hex = [];
+        for (let idx = source.length-1; idx >= 0; idx--) {
```
AndrejMitrovic, line 70: I think there should be an assert or check here somewhere, what if the length is zero?

src/modules/data/BlockHeader.ts
MichaelKim20: We have the hash value of the previous block.
I added this function to calculate the hash value of the current block and input it into the database.

AndrejMitrovic: Right you are correct.

We could have changed Agora to push a (Hash, Block) pair. But since you already did all this work then let's use your code.

AndrejMitrovic: The problem that I see is that every time we update the hashing mechanism in Agora we will also have to update it in Stoa. It means doubling the amount of necessary work.

MichaelKim20: Without this, we do not know the hash value of the last block entered.
I think it is necessary to develop other services such as Block Explorer.

AndrejMitrovic: Understandable. Thanks.

AndrejMitrovic:
Mostly LGTM. Can other people please review this too?

AndrejMitrovic (approved):

Geod24 (approved):

PR #98, opened 2020-08-25
Set the maximum number of blocks that can be recovered at one time

MichaelKim20:
Set the maximum number of blocks for stable operation of the system because trying to recover too many blocks at once requires a lot of resources in the system.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -35,6 +35,11 @@ class Stoa {
     protected task_manager: TaskManager;
 
     /**
+     * The maximum number of blocks that can be recovered at one time
+     */
+    private _max_count_on_recovery: number = 64;
+
+    /**
      * Constructor
      * @param database_filename sqlite3 database file name
      * @param agora_host The network host to connect to Agora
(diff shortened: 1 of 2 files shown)


src/Stoa.ts
```diff
@@ -34,6 +34,11 @@ class Stoa {
      */
     protected task_manager: TaskManager;
 
+    /**
+     * The maximum number of blocks that can be recovered at one time
+     */
+    private _max_count_recovery: number = 64;
```
```suggestion
    private _recovery_max_count: number = 64;
```
According to our naming convention, it is better for verbs to be defined first.

src/Stoa.ts
```diff
@@ -423,6 +436,22 @@ class Stoa {
...
+        return this._max_count_recovery;
+    }
+
+    /**
+     * Set the maximum number of blocks that can be recovered at one time
+     */
+    public set max_count_recovery (value: number)
+    {
+        this._max_count_recovery = value;
+    }
```
```suggestion
    get max_count_recovery (): number
    {
        return this._max_count_recovery;
    }

    /**
     * Set the maximum number of blocks that can be recovered at one time
     */
    set max_count_recovery (value: number)
    {
        this._max_count_recovery = value;
    }
```
I don't think `Getter/Setter` needs a `public`.

src/Stoa.ts
MichaelKim20: I shortened `**max count blocks on recovery**`.

MichaelKim20:
@TrustHenry I fixed it

TrustHenry (approved):

TrustHenry:
If the expected block height is higher even after the recovery to the maximum recovery height, is the next operation also guaranteed?

MichaelKim20:
> If the expected block height is higher even after the recovery to the maximum recovery height, is the next operation also guaranteed?

Yes, If the block height is less than the target height during the recovery, it is implemented to recover at the next cycle.

PR #100, opened 2020-09-11
Change data type of the height of the block and amount from number to UInt64

### LINKED ISSUE #82: Change data type of the height of the block and amount  from number to bigint.
We should either change to the same data type as Agora or find an alternative.

MichaelKim20:
The `Height` and `Amount` are a 64-byte unsigned integer. 
However, BigInt is a signed integer.

I found and added a library that supports 64-byte unsigned integer. 
And applied it to the code.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -118,16 +118,19 @@ class Stoa {
                         {
                             let preimage_hash: Buffer = row.preimage_hash;
                             let preimage_distance: number = row.preimage_distance;
-                            let target_height: number = row.height;
+                            let target_height: UInt64 = UInt64.fromNumber(row.height);
                             let result_preimage_hash = new Hash();
-                            let start_index: number = row.enrolled_at + 1;
+                            let start_index: UInt64 = UInt64.add(UInt64.fromNumber(row.enrolled_at), 1);
 
                             // Hashing preImage
-                            if ((target_height >= start_index) &&
-                                (start_index + row.preimage_distance) >= target_height)
+                            if (
+                                (UInt64.compare(target_height, start_index) >= 0) &&
+                                (UInt64.compare(UInt64.add(start_index, row.preimage_distance), target_height) >= 0)
+                            )
                             {
                                 result_preimage_hash.fromBinary(preimage_hash, Endian.Little);
-                                for (let i = 0; i < start_index + row.preimage_distance - target_height; i++)
+                                let count = Number(Utils.UInt64ToString(UInt64.sub(UInt64.add(start_index, row.preimage_distance), target_height)));
+                                for (let i = 0; i < count; i++)
                                 {
                                     result_preimage_hash = hash(result_preimage_hash.data);
                                     preimage_distance--;
@@ -146,11 +149,11 @@ class Stoa {
                                 } as IPreimage;
 
                             let validator: ValidatorData =
-                            new ValidatorData(row.address, row.enrolled_at,
-                                Hash.createFromBinary(row.stake, Endian.Little).toString(), preimage);
+                                new ValidatorData(row.address, UInt64.fromNumber(row.enrolled_at),
+                                    Hash.createFromBinary(row.stake, Endian.Little).toString(), preimage);
                             out_put.push(validator);
                         }
-                        res.status(200).send(JSON.stringify(out_put));
+                        res.status(200).send(Utils.toJson(out_put));
                     }
                     else
                     {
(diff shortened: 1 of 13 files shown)


tests/Utils.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import * as assert from 'assert';
+import { Utils } from "../src/modules/utils/Utils";
+
+describe ('Test of isNumber, isPositiveNumber, isNegativeNumber', () =>
```
TrustHenry: Recommend adding separator(,) and a decimal(.) test here.

src/modules/storage/LedgerStorage.ts
```diff
@@ -471,13 +473,13 @@ export class LedgerStorage extends Storages
...
                     [
-                        height,
+                        height.toString(),
                         tx_idx,
                         out_idx,
                         hash.toBinary(Endian.Little),
                         utxo_key.toBinary(Endian.Little),
                         output.address.toString(),
-                        output.value
+                        Utils.UInt64ToString(output.value)
```
TrustHenry, line 483: The Amount type has been treated as String.
I think it's good.
It would be good to provide an arithmetic function for the Amount type later.

src/modules/utils/Utils.ts
```diff
@@ -0,0 +1,104 @@
...
+        return /^\-([0-9]+)$/.test(value);
+    }
+
+    /**
+     * Converts the UInt64 to the string
+     * @param value The number of UInt64
+     * @param radix An integer in the range 2 through 36 specifying
+     * the base to use for representing numeric values.
+     */
+    public static UInt64ToString(value: UInt64, radix?: number): string
```
TrustHenry, line 51: Are you planning to add a reverse converter? (StringToUInt64)

tests/Utils.test.ts
MichaelKim20: We don't use floating-point at the momentarily.
And do not use, between numbers.
If it is used, how about adding it then?

src/modules/utils/Utils.ts
MichaelKim20: It is already embedded in UByte64.

tests/Utils.test.ts
TrustHenry: It's just adding a test code for a regular expression.
String can contain any character.

MichaelKim20: I don't think it will be easy to determine the exact number of decimal places, the location of commas, etc. with a regular expression.
So I renamed the function. isInteger, isPositiveInteger, isNegativeInteger

src/modules/utils/Utils.ts
```diff
@@ -67,4 +67,38 @@ export class Utils
...
+    /**
+     * It is a function that outputs an object with BigInt as JSON.
+     * @param data - Object to be converted to a JSON string
+     */
+    public static toJson(data: object): string
+    {
+        if (data === undefined)
+            return "";
+
+        let count1 = 0, count2 = 0;
```
AndrejMitrovic: This code is so unusual that I think it needs an explanation.

MichaelKim20: I added some document, Is this enough?

tests/Recovery.test.ts
```diff
@@ -124,14 +125,14 @@ class TestStoa extends Stoa
...
-                    Number.isNaN(req.query.block_height)
+                    !Utils.isPositiveInteger(req.query.block_height.toString())
                 )
                 {
                     res.status(204).send();
                     return;
                 }
 
-                let block_height = Math.max(Number(req.query.block_height), 0);
+                let block_height = new Height(UInt64.fromString(req.query.block_height.toString()));
```
MichaelKim20, line 135: The condition greater than zero has already been confirmed by Utils.isPositiveInteger.

src/Stoa.ts
```diff
@@ -323,35 +326,32 @@ class Stoa {
...
         {
             (async () => {
                 try
                 {
-                    let block_height = expected_height;
-                    let max_blocks = height - expected_height;
+                    let max_blocks = Number(UInt64.sub(height.value, expected_height.value).toString());
 
-                    if (max_blocks > this._max_count_on_recovery)
+                    if (UInt64.compare(max_blocks, this._max_count_on_recovery) > 0)
```
MichaelKim20, line 344: The source code of `UInt64.compare` is
https://github.com/princess-rosella/spu-integer-math/blob/master/lib/UInt64.ts#L185-L196

src/Stoa.ts
```diff
@@ -229,11 +235,11 @@ class Stoa {
                                 } as IPreimage;
 
                             let validator: ValidatorData =
-                                new ValidatorData(row.address, row.enrolled_at,
+                                new ValidatorData(row.address, UInt64.fromNumber(row.enrolled_at),
                                     Hash.createFromBinary(row.stake, Endian.Little).toString(), preimage);
                             out_put.push(validator);
                         }
-                        res.status(200).send(JSON.stringify(out_put));
+                        res.status(200).send(Utils.toJson(out_put));
```
MichaelKim20, line 242: `Utils.toJson` can convert objects containing UInt64 to JSON.

Geod24 (approved):

PR #102, opened 2020-09-15
Add descriptions of the function's parameters

MichaelKim20:
There was a missing description of the parameters.
I added descriptions.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -579,10 +579,11 @@ export class LedgerStorage extends Storages
     }
     /**
      * Get validators
-     * @param height: The height parameter is optional.
-     * validators based on the block height if there is a height.
-     * if height null the most up to date state is expected.
-     * if the height is null then current valid validators.
+     * @param height If present, the height at which the returned list of
+     * validators will apply. If absent, the highest height this stoa
+     * instance is aware of is assumed.
+     * @param address If present, only returns a single validator (or none if not enrolled).
+     * If null, all available validators are returned.
      * @returns Returns the Promise. If it is finished successfully the `.then`
      * of the returned Promise is called with the records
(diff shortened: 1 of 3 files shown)


Geod24 (approved):

src/modules/storage/LedgerStorage.ts
```diff
@@ -579,10 +579,11 @@ export class LedgerStorage extends Storages
     }
     /**
      * Get validators
-     * @param height: The height parameter is optional.
+     * @param height The height parameter is optional.
```
Geod24: A better way to explain it:
```suggestion
     * @param height If present, the height at which the returned list of validators will apply.
                       If absent, the highest height this stoa instance is aware of is assumed.
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -579,10 +579,11 @@ export class LedgerStorage extends Storages
...
      * validators based on the block height if there is a height.
      * if height null the most up to date state is expected.
      * if the height is null then current valid validators.
+     * @param address The address of the validator
```
Geod24: Likewise:
```suggestion
     * @param address If present, only returns a single validator (or none if not enrolled).
                                     If null, all available validators are returned.
```

If a parameter is optional, you have to describe what are the effect of providing it, or not.

src/modules/storage/Storages.ts
```diff
@@ -27,6 +27,9 @@ export class Storages
      * @param filename Valid values are filenames,
      * ":memory:" for an anonymous in-memory database and
      * an empty string for an anonymous disk-based database
+     * @param callback If provided, this function will be called when
+     * the database was opened successfully or when an error occurred.
+     * The first argument is an error object. If there is no error, this value is null.
```
Geod24, line 32: Perfect 👍

PR #104, opened 2020-09-23
Change the number to a string to preserve the precision of UInt64

### LINKED ISSUE #99: Add ability to maintain precision of UInt64 included in JSON
The data received by Http automatically convert from string to JSON internally.
We need to find a solution to maintain the precision of UInt64 in this conversion process.

MichaelKim20:
The data received by Http automatically convert from string to JSON internally.
I changed the number to a string to preserve the precision of UInt64 in JSON string

Related to #99


MichaelKim20:
Stoa's issue #99 was resolved as Agora's issue #1211(https://github.com/bpfkorea/agora/issues/1211) was resolved.
However, this problem can occur at any time because the `ulong` type can be used in Agora at any time.
Therefore, I request a review of this PR.

PR #107, opened 2020-09-24
Dockerfile: Revert to a single service Dockerfile

Geod24:
The original idea was to have a single docker container for simplicity.
As we're moving to a different setup, with servers and systemd unit,
we can remove this complication.


Dockerfile
```diff
@@ -1,10 +1,10 @@
...
-FROM bpfk/agora:latest
+FROM alpine:3.12
 RUN apk add --no-cache curl git python3 py-pip alpine-sdk \
-    bash autoconf libtool automake nodejs npm supervisor
+    bash autoconf libtool automake nodejs npm
 ADD . /stoa/
 WORKDIR /stoa/
 RUN npm ci
 EXPOSE 3836
 # Use the supervisord to run agora and stoa multiprocessor.
```
TrustHenry: A change in the content of the annotation is required.

TrustHenry (approved):

Dockerfile
Geod24: Fixed

Issue #106, opened 2020-09-24
Add a CLI option to pass the Agora address to contact


Geod24:
So one can do `docker run -it bpfk/stoa -- -agora=http://127.0.0.1:4826/`

Issue #85, opened 2020-08-14
Separate the various settings required for execution into a config file


MichaelKim20:
We can separate the settings using separate source code or JSON.
I think using YAML is also a good choice, considering that Stoa can work in Docker.

PR #113, opened 2020-10-07
Add an IP of a Docker Container to the whitelist of an IP Filter

### LINKED ISSUE #112: IP filtering of the stoa should be checked.
Could not access Stoa's endpoint '/block_externalized' and '/preimage_received' in the Docker Container.

MichaelKim20:
I added "::ffff:172.17.0.0/16".
After testing at the docker, I found that the block was well stored in the database.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -83,7 +83,7 @@ class Stoa {
         }, 10);
 
         // List that allows access only to those IPs
-        const white_ip_list: IpList = ['::ffff:127.0.0.1'];
+        const white_ip_list: IpList = ['::ffff:127.0.0.1','::ffff:172.17.0.0/16'];
 
         /**
          * Called when a request is received through the `/validators` handler


Geod24:
I'm not sure whitelisting IP will work very well. I would recommend two things instead:
- Ability to set the listening interface, so one can listen to either 127.0.0.1 or 0.0.0.0;
- Some sort of auth, e.g. HTTP basic auth;

The issue with whitelisting is that you still expose the port to the internet, it's only the app side that rejects connection.

TrustHenry:
> The issue with whitelisting is that you still expose the port to the internet, it's only the app side that rejects connection.

I think we need a physical firewall or an OS-level firewall to solve that problem.

> Some sort of auth, e.g. HTTP basic auth;

HTTP basic authentication is also a good idea.

Geod24:
> I think we need a physical firewall or an OS-level firewall to solve that problem.

Not necessarily. You can just set up an internal network and listen to this interface, if your machine has multiple interfaces, or if you set up a VPN.

TrustHenry:
We have an external endpoint that is open to the outside in the same port.

MichaelKim20:
> I'm not sure whitelisting IP will work very well. I would recommend two things instead:
> 
> * Ability to set the listening interface, so one can listen to either 127.0.0.1 or 0.0.0.0;
> * Some sort of auth, e.g. HTTP basic auth;
> 
> The issue with whitelisting is that you still expose the port to the internet, it's only the app side that rejects connection.

That's a good idea. I'll test it out.

MichaelKim20:
### **We have four endpoints in Stora.**
1. '/validators'
2. '/validator/:address'
3. '/block_externalized'
4. '/preimage_received'

### **This is the test result for the first proposal.**
I modified the Stoa server sockets to bind to '127.0.0.1'.
I could not access the endpoint of the stoa running in the Docker container.

Endpoint No. 1 and No. 2 are allowed access to all Internet addresses.
Endpoint No. 3 and No. 4 should only be accessible from Agora.
Endpoint, I think it will be difficult to solve it with the binding of server sockets in the environment that runs in the docker container.
I think it's a good idea at this stage to handle it so that it can be added to the whitelist of IP filtering.

### **On the second proposal.**

First, how to configure Stoa server with Agora should be clearly discussed.
There are two main cases.

**A. Endpoint No. 1 and No. 2 open to all users**

> 1. When operating without the use of a docker container on the same server.
a). In this case, binding the Stoa server sockets to '127.0.0.1' will allow access only locally.  
However, it is difficult to resolve with bindings on server sockets because 
endpoints No. 1 and No. 2 are not accessible.
b). I think the best alternative is IP filtering.
> 
> 2. When operating in a docker container on the same server
a). This issue is designed to be solved in this case.
b). I have confirmed that the binding of the server sockets cannot solve it.
c). This will also be resolved by adding to the whitelist of IP filtering.
> 3. Agora and Stoa operate on separate servers.
a). This will also be resolved by adding it to the whitelist of IP filtering.

**B. To ensure that all endpoints are accessible only to allowed users.**
> 1. There is no need for IP-filtering in this case.
> 2. In this case, we can proceed with @Geod24 's second proposal.
>  I think it is better to create and solve a separate issue.

This PR is to resolve Issue #106 which does not allow access to the Stoa endpoints No. 3 and No. 4 of the Docker Container.
Please review it again.

MichaelKim20:
I referred to the following document.
https://bluese05.tistory.com/15

src/Stoa.ts
```diff
@@ -83,7 +83,7 @@ class Stoa {
         }, 10);
 
         // List that allows access only to those IPs
-        const white_ip_list: IpList = ['::ffff:127.0.0.1'];
+        const white_ip_list: IpList = ['::ffff:127.0.0.1', "::ffff:172.17.0.0/16"];
```
TrustHenry: Unnecessary space behind code.

MichaelKim20: I fixed it

MichaelKim20:
I will close this PR because the issue has already been resolved by #85.

PR #114, opened 2020-10-07
Add .dockerignore for removing logs, docs, database, ...

### LINKED ISSUE #111: We need a `.dockerignore`
To ignore folders and files that are not needed to create a docker image.

MichaelKim20:
I added the list below to ignore folders and files that are not needed to create a docker image.
node_modules
docs
logs
database
.vscode
.github
.idea

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+# Part of the source tree, but not needed to build Stoa
+.git/
+.github/
+docs/
+logs/
+
+# Dependency build files, It is installed when creating a docker image.
+node_modules/
+
+# Database file
+database
+
+# Visual Studio Code
+.vscode/
+
+# WebStorm
+.idea/


MichaelKim20:
Below is a log that was recorded internally after the docker was executed.

[Stoa] 2020-10-07T04:32:52.849Z info Stoa API server listening at 3836

Geod24:
Need to add a trailing slash (`/`) otherwise 👍

Geod24:
> Below is a log that was recorded internally after the docker was executed.

I'm not sure I follow. The `.dockerignore` is only used when building the container, so a log of the run is a bit too late.
By the way, isn't `node_modules` required to run the app ?

MichaelKim20:
> > Below is a log that was recorded internally after the docker was executed.
> 
> I'm not sure I follow. The `.dockerignore` is only used when building the container, so a log of the run is a bit too late.
> By the way, isn't `node_modules` required to run the app ?

The node_modules will be reinstalled by the command below.
RUN npm ci

MichaelKim20:
> Need to add a trailing slash (`/`) otherwise 👍

 I fixed it

TrustHenry (approved):
Required when building a docker image, but good if you test it often.

MichaelKim20:
I made the following changes.
I also added .git.
```
# Part of the source tree, but not needed to build Stoa
.git/
.github/
docs/
logs/

# Dependency build files, It is installed when creating a docker image.
node_modules/

# Database file
database

# Visual Studio Code
.vscode/

# WebStorm
.idea/
```

Geod24 (approved):

PR #115, opened 2020-10-07
Implement the ability to receive the information needed to execute through the arguments

MichaelKim20:
We had previously tried to solve the problem using environmental variables.
I realized after several attempts that this was impossible.
I couldn't set the environment variable with 'npm start'.

Instead, I solved the problem in the following ways.

I implemented the ability to receive the information needed to execute through the arguments

Used as below.
npm start -- --agora=http://127.0.0.1:4826/  

Used in the docker as follows
docker run -d -p 3836:3836 -i -t stoa -- --agora=http://127.0.0.1:4826/

Additional arguments
'-a', '--agora' - The endpoint of Agora
'-p', '--port' - The network port of Stoa
'-d', '--database’ - The file name of sqlite3 database

Relates to #106

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@types/sqlite3": "^3.1.6",
     "ajv": "^6.12.3",
     "app-root-path": "^3.0.0",
+    "argparse": "^2.0.1",
     "assert": "^2.0.0",
     "axios": "^0.19.2",
     "body-parser": "^1.19.0",
@@ -47,6 +48,7 @@
     "winston": "^3.3.3"
   },
   "devDependencies": {
+    "@types/argparse": "^2.0.1",
     "@types/chai": "^4.2.12",
     "@types/crc": "^3.4.0",
     "@types/express": "^4.17.7",
--- a/package.json
+++ b/package.json
@@ -6,10 +6,10 @@
   "scripts": {
     "json2ts": "./node_modules/.bin/json2ts -i src/modules/data/schemas/ -o src/modules/data/types/",
     "_test": "mocha -r ts-node/register tests/**/*.test.ts",
-    "_start": "nodemon --watch src --delay 1 --exec 'ts-node' src/main.ts",
     "test": "NODE_ENV=test npm-run-all --serial json2ts _test",
     "debug": "NODE_ENV=development npm-run-all --serial json2ts _test",
-    "start": "NODE_ENV=production npm-run-all --serial json2ts _start",
+    "prestart": "npm run json2ts",
+    "start": "NODE_ENV=production nodemon -w src -d 1 -x ts-node src/main.ts",
     "tsc": "tsc"
   },
   "repository": {
--- a/src/main.ts
+++ b/src/main.ts
@@ -2,13 +2,27 @@ import 'source-map-support/register';
 import { logger } from './modules/common/Logger';
 import Stoa from './Stoa';
 
+import { ArgumentParser } from "argparse";
 import express from 'express';
 import { URL } from 'url';
 
-const port: number = Number(process.env.PORT) || 3836;
-const agora_address: URL = new URL(process.env.AGORA_ENDPOINT || "http://127.0.0.1:2826");
-// TODO: Replace hardcoded "127.0.0.1" with  agora_address.hostname (#106)
-const stoa: express.Application = new Stoa("database", "127.0.0.1", agora_address.port).stoa;
+// Parse the arguments
+const parser = new ArgumentParser();
+parser.add_argument('-a', '--agora', { help: "The endpoint of Agora" });
+parser.add_argument('--address', { help: "The address to which we bind to Stoa" });
+parser.add_argument('-p', '--port', { help: "The port on which we bind to Stoa" });
+parser.add_argument('-d', '--database', { help: "The file name of sqlite3 database" });
+let args = parser.parse_args();
 
-stoa.listen(port, () => logger.info(`Stoa API server listening at ${port}`))
+const address: string = args.address || "0.0.0.0";
+const port: number = Number(args.port || "3836");
+const agora_address: URL = new URL(args.agora || "http://127.0.0.1:2826");
+const database_filename: string = args.database || "database";
+logger.info(`Using Agora located at: ${agora_address.hostname}: ${agora_address.port}`);
+logger.info(`The address to which we bind to Stoa: ${address}`);
+logger.info(`The port to which we bind to Stoa: ${port}`);
+logger.info(`The file name of sqlite3 database: ${database_filename}`);
+const stoa: express.Application = new Stoa(database_filename, agora_address.hostname, agora_address.port).stoa;
+
+stoa.listen(port, address, () => logger.info(`Listening to requests on: ${address}:${port}`))
 .on('error', err => logger.error(err));
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,12 @@
       "resolved": "https://registry.npmjs.org/@types/app-root-path/-/app-root-path-1.2.4.tgz",
       "integrity": "sha1-p4twMoKzKsVN52j1US7MNWmRncc="
     },
+    "@types/argparse": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/@types/argparse/-/argparse-2.0.1.tgz",
+      "integrity": "sha512-C4jqpdasIkxgjHEfAtQEx59/rprJxP4R2GgPW6xdKeOSmEPf3jvQB706R7hebZAimKWPMU88l9MchWxtAcgT3g==",
+      "dev": true
+    },
     "@types/body-parser": {
       "version": "1.19.0",
       "resolved": "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.19.0.tgz",
@@ -293,12 +299,9 @@
       "dev": true
     },
     "argparse": {
-      "version": "1.0.10",
-      "resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz",
-      "integrity": "sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==",
-      "requires": {
-        "sprintf-js": "~1.0.2"
-      }
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz",
+      "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q=="
     },
     "array-filter": {
       "version": "1.0.0",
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -6,10 +6,9 @@
   "scripts": {
     "json2ts": "./node_modules/.bin/json2ts -i src/modules/data/schemas/ -o src/modules/data/types/",
     "_test": "mocha -r ts-node/register tests/**/*.test.ts",
-    "_start": "nodemon --watch src --delay 1 --exec 'ts-node' src/main.ts",
     "test": "NODE_ENV=test npm-run-all --serial json2ts _test",
     "debug": "NODE_ENV=development npm-run-all --serial json2ts _test",
-    "start": "NODE_ENV=production npm-run-all --serial json2ts _start",
```
MichaelKim20, line 12: npm-run-all
I removed npm-run-all from the contents of the npm start command.
This is because the arguments cannot be delivered in the individual command of npm-run-all.

package.json
```diff
@@ -6,10 +6,9 @@
...
+    "start": "NODE_ENV=production nodemon -w src -i src/modules/data/types -d 1 -x ts-node src/main.ts",
```
MichaelKim20: I changed to run `npm run json2ts` in `src/main.ts`.
The files in `src/modules/data/types` are changed.
And the service is restarted by nodemon.
In order to prevent the service from restarting, I made sure that the folder was ignored by the monitoring.

src/main.ts
```diff
@@ -2,13 +2,29 @@ import 'source-map-support/register';
...
+parser.add_argument('-a', '--agora', { help: "The endpoint of Agora" });
+parser.add_argument('-p', '--port', { help: "The network port of Stoa" });
+parser.add_argument('-d', '--database', { help: "The file name of sqlite3 database" });
+let args = parser.parse_args();
+
+const port: number = Number(args.port || "3836");
+const agora_address: URL = new URL(args.agora || "http://127.0.0.1:2826");
+const database_filename: string = args.database || "database";
+logger.info(`The network host to connect to Agora: ${agora_address.hostname}`);
+logger.info(`The network port to connect to Agora: ${agora_address.port}`);
```
TrustHenry: I think two lines of logs can be combined into one.

src/main.ts
```diff
@@ -2,13 +2,29 @@ import 'source-map-support/register';
...
-const agora_address: URL = new URL(process.env.AGORA_ENDPOINT || "http://127.0.0.1:2826");
-// TODO: Replace hardcoded "127.0.0.1" with  agora_address.hostname (#106)
-const stoa: express.Application = new Stoa("database", "127.0.0.1", agora_address.port).stoa;
+// Compile JSON Schema to TypeScript Type
+shell.exec("npm run json2ts");
+
+// Parse the arguments
+const parser = new ArgumentParser();
+parser.add_argument('-a', '--agora', { help: "The endpoint of Agora" });
+parser.add_argument('-p', '--port', { help: "The network port of Stoa" });
```
TrustHenry: This is good too.
It would be nice if I could parse this to the address.
It would be better to enter only one optional parameter.

src/main.ts
```diff
@@ -2,13 +2,29 @@ import 'source-map-support/register';
...
 import express from 'express';
+import shell from 'shelljs';
 import { URL } from 'url';
 
-const port: number = Number(process.env.PORT) || 3836;
-const agora_address: URL = new URL(process.env.AGORA_ENDPOINT || "http://127.0.0.1:2826");
-// TODO: Replace hardcoded "127.0.0.1" with  agora_address.hostname (#106)
-const stoa: express.Application = new Stoa("database", "127.0.0.1", agora_address.port).stoa;
+// Compile JSON Schema to TypeScript Type
+shell.exec("npm run json2ts");
```
TrustHenry: I don't know if we do typescript conversion here.
I think we can run it as a RUN on the Dockerfile.

MichaelKim20: That's a great idea.
We can run `npm run json2ts` when we create a docker image.
Thank you.

src/main.ts
MichaelKim20: I'd appreciate it if you could be more specific.

src/main.ts
```diff
@@ -2,13 +2,29 @@ import 'source-map-support/register';
...
+
+// Parse the arguments
+const parser = new ArgumentParser();
+parser.add_argument('-a', '--agora', { help: "The endpoint of Agora" });
+parser.add_argument('-p', '--port', { help: "The network port of Stoa" });
+parser.add_argument('-d', '--database', { help: "The file name of sqlite3 database" });
+let args = parser.parse_args();
+
+const port: number = Number(args.port || "3836");
+const agora_address: URL = new URL(args.agora || "http://127.0.0.1:2826");
```
TrustHenry, line 21: If you look at the code described, this value only supports pure address.
However, the default value includes ports.

src/main.ts
TrustHenry: For example, we would like to support `--agora http://127.0.0.1:2826` as the parameter address value.

src/main.ts
```diff
@@ -2,13 +2,29 @@ import 'source-map-support/register';
...
+const database_filename: string = args.database || "database";
+logger.info(`The network host to connect to Agora: ${agora_address.hostname}`);
+logger.info(`The network port to connect to Agora: ${agora_address.port}`);
+logger.info(`The network port of Stoa: ${port}`);
+logger.info(`The file name of sqlite3 database: ${database_filename}`);
+const stoa: express.Application = new Stoa(database_filename, agora_address.hostname, agora_address.port).stoa;
```
TrustHenry, line 27: This is not the port value received as a parameter variable.

src/main.ts
MichaelKim20: So I tested it after I applied it.
I have confirmed that it works well in the docker environment.
But it could be a problem when we run it on a local computer, not a docker environment.
So I think the current method would be good, what do you think?

src/main.ts
TrustHenry: Ah, this seems so right.

src/main.ts
MichaelKim20: I found a way to use `npm command prestart` instead of `shell.exec`. I solved it with Henry at IRL.
And applied it.

src/main.ts
```diff
@@ -2,13 +2,25 @@ import 'source-map-support/register';
...
-const stoa: express.Application = new Stoa("database", "127.0.0.1", agora_address.port).stoa;
+// Parse the arguments
+const parser = new ArgumentParser();
+parser.add_argument('-a', '--agora', { help: "The endpoint of Agora" });
+parser.add_argument('-p', '--port', { help: "The network port of Stoa" });
+parser.add_argument('-d', '--database', { help: "The file name of sqlite3 database" });
+let args = parser.parse_args();
+
+const port: number = Number(args.port || "3836");
+const agora_address: URL = new URL(args.agora || "http://127.0.0.1:2826");
```
TrustHenry, line 19: If the address is only in agora_address, port exception processing will be required.
For example, if `-a http://127.0.0.1`
In this case, we should apply the default port value.

package.json
```diff
@@ -6,10 +6,10 @@
   "scripts": {
     "json2ts": "./node_modules/.bin/json2ts -i src/modules/data/schemas/ -o src/modules/data/types/",
     "_test": "mocha -r ts-node/register tests/**/*.test.ts",
-    "_start": "nodemon --watch src --delay 1 --exec 'ts-node' src/main.ts",
     "test": "NODE_ENV=test npm-run-all --serial json2ts _test",
     "debug": "NODE_ENV=development npm-run-all --serial json2ts _test",
-    "start": "NODE_ENV=production npm-run-all --serial json2ts _start",
+    "prestart": "npm run json2ts",
```
TrustHenry, line 11: Nice!

src/main.ts
MichaelKim20: If the port is not set, it is being treated as the http default of 80.

MichaelKim20: Agora's connection information will be resolved by proceeding with Issue #85 to use endpoints without separating the host from the port.

TrustHenry (approved):
LGTM

MichaelKim20:
### **We have four endpoints in Stora.**
1. '/validators'
2. '/validator/:address'
3. '/block_externalized'
4. '/preimage_received'

### **This is the test result for the first proposal.**
I modified the Stoa server sockets to bind to '127.0.0.1'.
I could not access the endpoint of the stoa running in the Docker container.

Endpoint No. 1 and No. 2 are allowed access to all Internet addresses.
Endpoint No. 3 and No. 4 should only be accessible from Agora.
Endpoint, I think it will be difficult to solve it with the binding of server sockets in the environment that runs in the docker container.
I think it's a good idea at this stage to handle it so that it can be added to the whitelist of IP filtering.

### **On the second proposal.**

First, how to configure Stoa server with Agora should be clearly discussed.
There are two main cases.

**A. Endpoint No. 1 and No. 2 open to all users**

> 1. When operating without the use of a docker container on the same server.
a). In this case, binding the Stoa server sockets to '127.0.0.1' will allow access only locally.  
However, it is difficult to resolve with bindings on server sockets because 
endpoints No. 1 and No. 2 are not accessible.
b). I think the best alternative is IP filtering.
> 
> 2. When operating in a docker container on the same server
a). This issue is designed to be solved in this case.
b). I have confirmed that the binding of the server sockets cannot solve it.
c). This will also be resolved by adding to the whitelist of IP filtering.
> 3. Agora and Stoa operate on separate servers.
a). This will also be resolved by adding it to the whitelist of IP filtering.

**B. To ensure that all endpoints are accessible only to allowed users.**
> 1. There is no need for IP-filtering in this case.
> 2. In this case, we can proceed with @Geod24 's second proposal.
>  I think it is better to create and solve a separate issue.

This PR is to resolve Issue #106 which does not allow access to the Stoa endpoints No. 3 and No. 4 of the Docker Container.
Please review it again.

src/main.ts
```diff
@@ -2,13 +2,25 @@ import 'source-map-support/register';
...
+const database_filename: string = args.database || "database";
+logger.info(`The network host to connect to Agora: ${agora_address.hostname}`);
+logger.info(`The network port to connect to Agora: ${agora_address.port}`);
```
Geod24:
```suggestion
logger.info(`Using Agora located at: ${agora_address.hostname}: ${agora_address.port}`);
```

src/main.ts
```diff
@@ -2,13 +2,25 @@ import 'source-map-support/register';
...
+logger.info(`The network port of Stoa: ${port}`);
```
Geod24:
```suggestion
logger.info(`Listening to requests on: ${address}:${port}`);
```

Of course `${address}` needs to be implemented

Geod24 (approved):
Nice, just two comments

src/main.ts
MichaelKim20: I fixed it as below according to the review.
```TypeScript
    const address: string = args.address || "0.0.0.0";
    const port: number = Number(args.port || "3836");
    const agora_address: URL = new URL(args.agora || "http://127.0.0.1:2826");
    const database_filename: string = args.database || "database";
    logger.info(`Using Agora located at: ${agora_address.hostname}: ${agora_address.port}`);
    logger.info(`Listening to requests on: ${address}:${port}`);
    logger.info(`The file name of sqlite3 database: ${database_filename}`);
    const stoa: express.Application = new Stoa(database_filename, agora_address.hostname, agora_address.port).stoa;

    stoa.listen(port, address, () => logger.info(`Stoa API server listening at ${port}`))
    .on('error', err => logger.error(err));
```

src/main.ts
```diff
@@ -2,13 +2,26 @@ import 'source-map-support/register';
...
+parser.add_argument('-d', '--database', { help: "The file name of sqlite3 database" });
+let args = parser.parse_args();
 
-stoa.listen(port, () => logger.info(`Stoa API server listening at ${port}`))
+const address: string = args.address || "0.0.0.0";
+const port: number = Number(args.port || "3836");
+const agora_address: URL = new URL(args.agora || "http://127.0.0.1:2826");
+const database_filename: string = args.database || "database";
+logger.info(`Using Agora located at: ${agora_address.hostname}: ${agora_address.port}`);
+logger.info(`Listening to requests on: ${address}:${port}`);
```
Geod24: Ah this is already done at L26, so could you just replace the log message at L26 with this ?

MichaelKim20: I fixed it

PR #116, opened 2020-10-08
Implement the ability to parse the YAML-format configuration

### LINKED ISSUE #85: Separate the various settings required for execution into a config file

MichaelKim20:
I Implemented the ability to parse the YAML-format configuration.
I applied the class `Config` to `main.ts`.

I am planning to apply Agumert after issue 106 is resolved.
Also, I have a plan to apply it to logging and database.
Also, I plan to make the shared directory of the docker easier to use.

Relates to issue #85


package.json
```diff
@@ -38,7 +38,10 @@
     "crc": "^3.8.0",
     "express": "^4.17.1",
     "express-ipfilter": "^1.1.2",
+    "extend": "^3.0.2",
+    "js-yaml": "^3.14.0",
     "json-schema-to-typescript": "^9.1.1",
+    "mkdirp": "^1.0.4",
```
MichaelKim20, line 44: Like mkdir -p
It creates a new directory and any necessary subdirectories at dir.

See https://www.npmjs.com/package/mkdirp

package.json
```diff
@@ -38,7 +38,10 @@
     "crc": "^3.8.0",
     "express": "^4.17.1",
     "express-ipfilter": "^1.1.2",
+    "extend": "^3.0.2",
```
MichaelKim20, line 41: Extend one object with one or more others, returning the modified object.
``` JavaScript
var extend = require('extend');
extend(targetObject, object1, object2);
```
See https://www.npmjs.com/package/extend

package.json
```diff
@@ -38,7 +38,10 @@
...
+    "js-yaml": "^3.14.0",
```
MichaelKim20, line 42: See https://www.npmjs.com/package/js-yaml

MichaelKim20:
As discussed in PR #118 , I changed the endpoint of Agora to a URL object.
This changed the ServerConfig.agora_endpoint within Config.ts to an object in the URL.

tests/config.example.yaml
```diff
@@ -0,0 +1,32 @@
...
+################################################################################
+database:
+  filename: APP_ROOT/stoa/data/database
+
+################################################################################
+##                               Logging options                              ##
+################################################################################
+logging:
+  # The name of the folder where the logs are stored.
+  folder: APP_ROOT/stoa/logs/
```
Geod24: Instead of building a templating language here, it would be better to just make it relative to the working directory.

MichaelKim20: Okay. I'll try to fix it.

MichaelKim20: @Geod24 I fixed it according to your review.

Geod24:
What is `appRootPath` now ? Is it the path to the binary ?

MichaelKim20:
> What is `appRootPath` now ? Is it the path to the binary ?

appRootPath is an npm module called 'app-root-path' and is defined as follows:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/app-root-path/index.d.ts#L6-L37

This also has a resolve function, but it did not work the same as path.resolve, so it was not available.

path.resolve selects the second factor if the second factor is the absolute path.
However, appRootPath just returns the combined results.

So I used appRootPath.toString().

Geod24:
What I mean is that this module doesn't seem required, since all path should be relative to the CWD. From the documentation, it seems to be mostly useful for its resolve function.

Regardless, let's move ahead with this. It just needs a rebase.

PR #118, opened 2020-10-13
Change the host and port of Agora to the endpoint

MichaelKim20:
Agora's information is entered via the command line parameter in the endpoint rather than the hostname and port (ex http://127.0.0.1:2826).

Inside the store, it was used separately from the hostname and port.
Therefore, it will not function normally when the protocol is changed.

I changed this so that Agora's endpoint was used as they were inside Stoa. Thus, it may be possible to apply the protocol that may change in the future.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -13,6 +13,7 @@ import cors from 'cors';
 import express from 'express';
 import { IpFilter, IpList } from 'express-ipfilter';
 import { UInt64 } from 'spu-integer-math';
+import { URL } from 'url';
 
 class Stoa {
     public stoa: express.Application;
@@ -20,14 +21,9 @@ class Stoa {
     public ledger_storage: LedgerStorage;
 
     /**
-     * The network host to connect to Agora
+     * The network endpoint to connect to Agora
      */
-    private readonly agora_host: string;
-
-    /**
-     * The network port to connect to Agora
-     */
-    private readonly agora_port: string;
+    private readonly agora_endpoint: URL;
 
     /**
      * The temporary storage of the data received from Agora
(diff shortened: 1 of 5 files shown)


TrustHenry:
LGTM
However, I think `agora_address` is more appropriate than `agora_endpoint`.

Geod24:
Why not pass an URL instead ?

MichaelKim20:
> agora_address

How about agora_url?

> LGTM
> However, I think `agora_address` is more appropriate than `agora_endpoint`.

How about agora_url?

MichaelKim20:
> Why not pass an URL instead ?

If @Geod24 's  suggestion is to convey URL objects created using 'npm module url', that is a good idea.
I intended to simplify coding in 'main.ts'.
Therefore, I tried to organize the parameters with only strings and numbers here.
Internally, AgoraClient uses URL objects to complete the endpoints and request data from Agora.
Therefore, it would be natural to handle it with a string here.
Thank you for your opinion.

Geod24:
Yes, the idea is to use the URL module. Because `URL` is a structured representation, while strings are not structured.
It's almost always better to pass an `URL` object than a string (host) + short (port) + string (path).
Also, regarding the naming, I'm personally happy with `endpoint`.

MichaelKim20:
> Yes, the idea is to use the URL module. Because `URL` is a structured representation, while strings are not structured.
> It's almost always better to pass an `URL` object than a string (host) + short (port) + string (path).
> Also, regarding the naming, I'm personally happy with `endpoint`.

@Geod24 Thank you for your review 
I fixed it. I will apply this to other PR.

Geod24 (approved):

PR #127, opened 2020-10-20
Add the ability to validate the IP address

### LINKED ISSUE #122: Set default value of address to bind to to `127.0.0.1`
```
[Stoa] 2020-10-20T08:46:04.494Z info Using Agora located at: http://127.0.0.1:2826/
[Stoa] 2020-10-20T08:46:04.496Z info The address to which we bind to Stoa:
[Stoa] 2020-10-20T08:46:04.496Z info The port to which we bind to Stoa: 3836
[Stoa] 2020-10-20T08:46:04.496Z info The file name of sqlite3 database: /stoa/bin/stoa/data/Stoa.db
[Stoa] 2020-10-20T08:46:04.511Z info Listening to requests on: :3836
```

A few things:
- "The address to which [...]" and "The port to which [...]" are redundant and can be removed
- We should check the validity of the address
- We should listen to `127.0.0.1` by default, not the empty string.

MichaelKim20:
I set default value of address to bind to from `` to `127.0.0.1`.

from issue #122 
```
The address to which [...]" and "The port to which [...]" are redundant and can be removed
We should check the validity of the address
We should listen to 127.0.0.1 by default, not the empty string.
```


Geod24:
Could you also add validation so it rejects empty or invalid values ? At the moment, what would happen if someone wrote, say, "helloworld" as the address ?

MichaelKim20:
> Could you also add validation so it rejects empty or invalid values ? At the moment, what would happen if someone wrote, say, "helloworld" as the address ?

I applied a class `IPValidator` that can validate the IP address.

Geod24:
Can't we use the URL type for validation ?

MichaelKim20:
I added a validation of the IP address. and I changed default value of address to "0.0.0.0"

src/modules/common/IPValidator.ts
```diff
@@ -0,0 +1,23 @@
...
+export class IPValidator
+{
+    public static check (ip: string): boolean
+    {
+        if (/^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/.test(ip))
+            return true;
+        else
+            return /(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))/.test(ip);
+    }
+}
```
Geod24: There's no builtin way in JS to verify an IP, such as `URL` ?

MichaelKim20: I found this, but not supported for TypeScript

MichaelKim20: https://www.npmjs.com/package/ip-validator

MichaelKim20: https://www.npmjs.com/package/ip 
I think this is possible in both JavaScript and TypeScript. How about this?

MichaelKim20: I applied npm module 'ip'

MichaelKim20:
Because IP addresses can be changed using the configuration file, I developed them intensively to verify IP addresses.

src/modules/common/Config.ts
```diff
@@ -136,9 +137,13 @@ export class ServerConfig implements IServerConfig
...
     {
-        let conf = extend(true, {}, ServerConfig.defaultValue());
+        let default_config = ServerConfig.defaultValue();
+        let conf = extend(true, {}, default_config);
         extend(true, conf, {address: address, port: port, agora_endpoint: agora_endpoint});
-        this.address = conf.address;
+        if (ip.isV4Format(conf.address) || ip.isV6Format(conf.address))
+            this.address = conf.address;
+        else
+            this.address = default_config.address;
```
Geod24: No! If the user enters wrong data, we *need* to inform them, no silently default to something else.
Having sensible defaults is important, but we should, *never*, *ever*, silently ignore wrong input from the user, it just creates an absolutely horrible user experience.

MichaelKim20: Yeah, that's right. That's a good idea.
I think I want automatic too much.

MichaelKim20: When the wrong IP address was entered, I changed the process to output the log and exit.

Geod24 (approved):

Issue #121, opened 2020-10-20
Make all path relatives to the current working directory


Geod24:
At the moment Stoa uses `appRoot` in many places. This creates issues while deploying with `docker`, as it stores the data and logs in a position relative to the binary which cannot be easily overriden.
Instead, `appRoot` shouldn't be used, and relative path should be resolved relative to the working directory.

MichaelKim20:
The places where `appRoot` is used were all used as below.
path.resolve(appRoot.toString(), 'B')
If B is an absolute path starting with "/", the value of B is used. 
ex) /stoa/stoa/database => /stoa/stoa/database
However, without "/", the appRoot and B are combined.
ex) stoa/database   => /stoa/stoa/database

Geod24:
`appRoot` gives the application directory.
The content of this repository is copied to `/stoa/bin/` in the docker image, so `appRoot` turns out to be `/stoa/bin/stoa/`.
In the live system, we mount `/srv/stoa` from the server to `/stoa/wd/` (working directory).
The config file is stored in `/srv/stoa/config.yaml`, and as a result is available as `/stoa/wd/config.yaml`.
So we're starting the image with `docker run <img> -- -c /stoa/wd/config.yaml`, and want `data` and `logs` to be written to `/stoa/wd/{data,logs}` so that if we stop and restart stoa, the data persists, and can be inspected from the host.

MichaelKim20:
Nodes exist in multiple directories on multiple source files. Because the location of the current folder is different for each file, a base point is required. So it was developed based on the root of the app.
Therefore, I don't think this is a bug in the program.
I think this issue is to change the way we operate.
And I think it is good to use cwd.And I think it is good to use cwd.
The same goes for this, and I will save cwd at the program's starting point and change it to use it by referring to it.
I want the type-bug removed.

Geod24:
I can see `appRoot` used here:
- https://github.com/bpfkorea/stoa/blob/91e6b1b66ad43c6e693fcac48acd57494812ce7e/src/modules/common/Logger.ts#L29 to set the log directory
- https://github.com/bpfkorea/stoa/blob/91e6b1b66ad43c6e693fcac48acd57494812ce7e/src/modules/common/Config.ts#L65 to read the config file
- https://github.com/bpfkorea/stoa/blob/91e6b1b66ad43c6e693fcac48acd57494812ce7e/src/modules/common/Config.ts#L100 to set the path of the database
[...]

PR #130, opened 2020-10-21
Change all paths from app-root to working directory

MichaelKim20:
I use `process.env.INIT_CWD`, If this value is undefined, use the root path of the app.

Relates to #121

### DIFF
--- a/src/modules/utils/Utils.ts
+++ b/src/modules/utils/Utils.ts
@@ -114,14 +114,13 @@ export class Utils
 
     /**
      *  Gets the path to where the execution command was entered for this process.
-     *  If this value is not set, returns the root path of the app.
+     * This value must be set, otherwise the application will terminate.
      */
     public static getInitCWD (): string
     {
-        //  A type of process.env.INIT_CWD is string | undefined.
-        //  In order to use this, it is always necessary to make sure
-        //  that it is `undefined`.
-        //  `string` can be return only when it is not `undefined`.
+        //  The type of `process.env.INIT_CWD` is `string | undefined`.
+        //  In order to simply return `string`, it is necessary to make sure
+        //  that it is not `undefined` first.
         if (process.env.INIT_CWD === undefined)
             assert.fail("INIT_CWD is not defined.");
         else
--- a/src/modules/utils/Utils.ts
+++ b/src/modules/utils/Utils.ts
@@ -11,6 +11,7 @@
 
 *******************************************************************************/
 
+import * as assert from 'assert';
 import { UInt64 } from 'spu-integer-math';
 
 export class Utils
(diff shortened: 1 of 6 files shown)


MichaelKim20:
If this value is not defined, I will have to think more about what value to use.

Geod24:
> If this value is not defined, I will have to think more about what value to use.

We should `assert` if it is not defined.

src/modules/utils/Utils.ts
```diff
@@ -110,4 +111,20 @@ export class Utils
 
         return res;
     }
+
+    /**
+     *  Gets the path to where the execution command was entered for this process.
+     *  If this value is not set, returns the root path of the app.
```
Geod24:
```suggestion
     * This value must be set, otherwise the application will terminate.
```

src/modules/utils/Utils.ts
```diff
@@ -110,4 +111,20 @@ export class Utils
...
+     */
+    public static getInitCWD (): string
+    {
+        //  A type of process.env.INIT_CWD is string | undefined.
+        //  In order to use this, it is always necessary to make sure
+        //  that it is `undefined`.
+        //  `string` can be return only when it is not `undefined`.
+        if (process.env.INIT_CWD === undefined)
```
MichaelKim20, line 124: Using 'process.env.INIT_CWD' found by @Geod24 , the working directory could be found.

src/modules/utils/Utils.ts
```diff
@@ -110,4 +111,20 @@ export class Utils
...
+            assert.fail("INIT_CWD is not defined.");
```
MichaelKim20, line 125: I used `assert` as suggested by @Geod24 .

MichaelKim20 (commented):
When running on the docker, I checked that the log and database were stored in the working directory.

src/modules/utils/Utils.ts
```diff
@@ -110,4 +111,20 @@ export class Utils
...
+    /**
+     *  Gets the path to where the execution command was entered for this process.
+     *  If this value is not set, returns the root path of the app.
+     */
+    public static getInitCWD (): string
+    {
+        //  A type of process.env.INIT_CWD is string | undefined.
+        //  In order to use this, it is always necessary to make sure
+        //  that it is `undefined`.
+        //  `string` can be return only when it is not `undefined`.
```
Geod24:
```suggestion
        //  The type of `process.env.INIT_CWD` is `string | undefined`.
        //  In order to simply return `string`, it is necessary to make sure
        //  that it is not `undefined` first.
```

Geod24 (approved):
Just two minor edits, otherwise LGTM.

PR #141, opened 2020-10-23
Simplify CI and extend to Node.js 15.x

Geod24:


Geod24:
>      AssertionError [ERR_ASSERTION]: INIT_CWD is not defined.

Seeing it locally as well (hence why I raised the PR). Will look into it.

Geod24:
Okay, I'm using `PWD` as a workaround, and it's working. And here I thought using Node.JS would be less buggy.

PR #142, opened 2020-10-26
Collection of minor improvements to the code

Geod24:
Working on fixing then numerous bugs we found in production, ended up with a bunch of simple commits.

### DIFF
--- a/src/modules/data/Height.ts
+++ b/src/modules/data/Height.ts
@@ -30,12 +30,12 @@ export class Height
      * Construct
      * @param value - The block height
      */
-    constructor (value?: UInt64)
+    constructor (value: bigint | string | UInt64)
     {
-        if (value !== undefined)
+        if (value instanceof UInt64)
             this.value = new UInt64(value);
         else
-            this.value = UInt64.fromNumber(0);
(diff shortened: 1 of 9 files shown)


src/modules/data/Height.ts
```diff
@@ -30,12 +30,11 @@ export class Height
      * Construct
      * @param value - The block height
      */
-    constructor (value?: UInt64)
+    constructor (value: bigint);
+    constructor (value: UInt64);
+    constructor (value: any)
```
MichaelKim20: How about this?
```suggestion
    constructor (value: bigint | string | UInt64)
    {
        if (typeof value == "bigint")
        {
            this.value = UInt64.fromString(value.toString());
        }
        else if (typeof value === "string")
        {
            this.value = UInt64.fromString(value);
        }
        else
        {
            this.value = new UInt64(value);
        }
    }
```

Geod24: Good suggestion, I committed a slightly modified version.

Geod24:
Should be fixed now.

Geod24:
Hum, shall we drop support for 10.x and 12.x ?

MichaelKim20 (approved):

PR #145, opened 2020-10-27
LedgerStorage: Only rollback after begin()

Geod24:
A common mistake which was also done in Agora.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -164,11 +164,11 @@ export class LedgerStorage extends Storages
         return new Promise<void>((resolve, reject) =>
         {
             (async () => {
+                let block: Block = new Block();
+                block.parseJSON(data);
+                await this.begin();
                 try
                 {
-                    let block: Block = new Block();
-                    block.parseJSON(data);
-                    await this.begin();
                     await saveBlock(this, block);
                     await this.putTransactions(block);
                     await this.putEnrollments(block);


src/modules/storage/LedgerStorage.ts
```diff
@@ -164,11 +164,11 @@ export class LedgerStorage extends Storages
         return new Promise<void>((resolve, reject) =>
         {
             (async () => {
+                let block: Block = new Block();
+                block.parseJSON(data);
+                await this.begin();
```
MichaelKim20, line 169: `block.parseJSON(data);` and `await this.begin()` 
Exception may occur here as well.
`reject()` should be called when an exception occurs here.

TrustHenry: Yes, I think so, too.
I think `let block: Block = new Block(); block.parseJSON(data);` can be move in `try`.

MichaelKim20: I have confirmed that when an exception occurs in the code below, it is not handled normally.
``` TypeScript
        function testThrowError (value: number): Promise<number>
        {
            return new Promise<number>((resolve, reject) => {
                if (value === 0)
                    throw new Error("error");

                resolve(value*value);
            })
        }

        try
        {
            (async () =>
            {
                await testThrowError(0);
            })();
        }
        catch (e)
        {
            console.log(e);
        }
```

When I changed it as follows, I checked that it was handled normally.
``` TypeScript
        (async () =>
        {
            try
            {
                await testThrowError(0);
            }
            catch (e)
            {
                console.log(e);
            }
        })();
```

Geod24:
Replaced by #163

PR #146, opened 2020-10-27
Simplify construction of `Height`

Geod24:
```
Since the constructor accepts `bigint | string | UInt64`,
there's no need to call either `fromString` or `fromNumber`,
reducing the dependencies between modules.
```

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -115,7 +115,7 @@ class Stoa extends WebService
         }
 
         let height = (req.query.height !== undefined)
-            ? new Height(UInt64.fromString(req.query.height.toString()))
+            ? new Height(req.query.height.toString())
             : null;
 
         if (height != null)
@@ -199,7 +199,7 @@ class Stoa extends WebService
         }
 
         let height = (req.query.height !== undefined)
-            ? new Height(UInt64.fromString(req.query.height.toString()))
+            ? new Height(req.query.height.toString())
             : null;
 
         let address: string = String(req.params.address);
@@ -333,7 +333,7 @@ class Stoa extends WebService
             throw Error("Not found block height in JSON Block");
         }
 
-        return new Height(UInt64.fromString(block.header.height));
+        return new Height(block.header.height);
     };
 
     /**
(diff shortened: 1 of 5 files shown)


MichaelKim20 (approved):
LGTM

src/Stoa.ts
```diff
@@ -115,7 +115,7 @@ class Stoa extends WebService
         }
 
         let height = (req.query.height !== undefined)
-            ? new Height(UInt64.fromString(req.query.height.toString()))
```
TrustHenry, line 118: https://github.com/bpfkorea/stoa/blob/7e2e8e0fef1cecbc0ceed2a35e66311792dffb8d/src/modules/data/Height.ts#L38
Two toString() runs.

src/Stoa.ts
```diff
@@ -115,7 +115,7 @@ class Stoa extends WebService
...
+            ? new Height(req.query.height.toString())
```
TrustHenry, line 118: https://github.com/bpfkorea/stoa/blob/7e2e8e0fef1cecbc0ceed2a35e66311792dffb8d/src/modules/data/Height.ts#L33-L39
I think we need to change the Height.
The string type is entered.
two toString() runs

src/Stoa.ts
```diff
@@ -199,7 +199,7 @@ class Stoa extends WebService
...
```
TrustHenry, line 202: same.

tests/Recovery.test.ts
```diff
@@ -124,7 +124,7 @@ class TestStoa extends Stoa
                     return;
                 }
 
-                let block_height = new Height(UInt64.fromString(req.query.block_height.toString()));
+                let block_height = new Height(req.query.block_height.toString());
```
TrustHenry, line 127: same.

Geod24:
> The string type is entered.
> two toString() runs

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toString
`String.prototype.toString` is equivalent to `valueOf`, so it's a no-op.

E.g. try:
```console
$ node -e 'console.info("hello".toString().toString().toString());'
```

And you'll get the same result no matter how many `toString()` you add.

PR #147, opened 2020-10-27
Add `await` in front of `assert.rejects`

MichaelKim20:
`asset.rejects` return Promise, so when using async, must put await before it.

### DIFF
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -245,7 +245,7 @@ describe ('Test for storing block data in the database', () =>
         block.parseJSON(sample_data0);
 
         await ledger_storage.putTransactions(block);
-        assert.rejects(ledger_storage.putTransactions(block),
+        await assert.rejects(ledger_storage.putTransactions(block),
             {
                 message: "SQLITE_CONSTRAINT: UNIQUE constraint failed:" +
                     " transactions.block_height, transactions.tx_index"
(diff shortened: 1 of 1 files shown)


Geod24:
Another alternative is to use `assert.throws` which is synchronous.

MichaelKim20:
> Another alternative is to use `assert.throws` which is synchronous.

That's a good idea.
So I tried to change it according to the proposal.
I thought it would be good to use a function that does not return Promise.

TrustHenry (approved):
I tested Node v15.0.1
and npm update  && npm test all pass
```
  "dependencies": {
    "@ctrl/ts-base32": "^1.2.3",
    "@types/app-root-path": "^1.2.4",
    "@types/cors": "^2.8.7",
    "@types/sodium-native": "^2.3.5",
    "@types/sqlite3": "^3.1.6",
    "ajv": "^6.12.3",
    "argparse": "^2.0.1",
    "assert": "^2.0.0",
    "axios": "^0.19.2",
    "body-parser": "^1.19.0",
    "cors": "^2.8.5",
    "crc": "^3.8.0",
    "express": "^4.17.1",
    "extend": "^3.0.2",
    "js-yaml": "^3.14.0",
    "json-schema-to-typescript": "^9.1.1",
    "mkdirp": "^1.0.4",
    "smart-buffer": "^4.1.0",
    "sodium-native": "^3.2.0",
    "source-map-support": "^0.5.19",
    "spu-integer-math": "0.0.3",
    "sqlite3": "^4.2.0",
    "urijs": "^1.19.2",
    "winston": "^3.3.3"
  }
```

Geod24 (approved):

PR #148, opened 2020-10-27
Change err?. to a more stable syntax

MichaelKim20:
The following error is occurring:
Therefore, I change to a more stable syntax.

```
/tests/Storage.test.ts:45
            assert.ok(!err, err?.message);
                                           ^

SyntaxError: Unexpected token '.'
```

### DIFF
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -27,7 +27,8 @@ describe ('Test ledger storage and inquiry function.', () =>
     {
         ledger_storage = new LedgerStorage(":memory:", (err: Error | null) =>
         {
-            assert.ok(!err, err?.message);
+            if (err)
+                assert.fail(err.message);
             doneIt();
         });
     });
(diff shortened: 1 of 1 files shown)


TrustHenry:
LGTM
I saw the same error. However, the error disappeared after the npm update.

TrustHenry:
A null check is required.

MichaelKim20:
> A null check is required.

I think this is a null test.
```
if (err)
```

TrustHenry:
> > A null check is required.
> 
> 
> 
> I think this is a null test.
> 
> ```
> 
> if (err)
> 
> ```
> 
> 

If it is null, there will be no error.

MichaelKim20:
The following statement is executed when `err` is not `null`.

MichaelKim20:
`if (err)` is not present, an error will be displayed in the following line and an error will be generated on execution. TypeScript clearly indicates

TrustHenry (approved):

PR #149, opened 2020-10-28
Improve serialization of Height and add BigInt serialization

Geod24:
This is a prelude to completely switching over to BigInt.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -15,6 +15,13 @@ import express from 'express';
 import { UInt64 } from 'spu-integer-math';
 import { URL } from 'url';
 
+// Module extension to allow customizing JSON serialization
+declare global {
+    interface BigInt {
+        toJSON(key?: string): string;
+    }
+}
+
 class Stoa extends WebService
 {
     public ledger_storage: LedgerStorage;
@@ -69,6 +76,11 @@ class Stoa extends WebService
         {
             return this.task();
         }, 10);
+
+        // Allow JSON serialization of BigInt
+        BigInt.prototype.toJSON = function(key?: string) {
+            return this.toString();
+        }
     }
 
     /**
--- a/src/modules/data/Height.ts
+++ b/src/modules/data/Height.ts
@@ -64,4 +64,15 @@ export class Height
     {
         return Utils.UInt64ToString(this.value);
     }
+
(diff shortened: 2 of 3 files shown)


src/Stoa.ts
```diff
@@ -15,6 +15,13 @@ import express from 'express';
 import { UInt64 } from 'spu-integer-math';
 import { URL } from 'url';
 
+// Module extension to allow customizing JSON serialization
+declare global {
+    interface BigInt {
+        toJSON(key? : string): string;
```
```suggestion
        toJSON(key?: string): string;
```

src/Stoa.ts
```diff
@@ -69,6 +76,11 @@ class Stoa extends WebService
         {
             return this.task();
         }, 10);
+
+        // Allow JSON serialization of BigInt
+        BigInt.prototype.toJSON = function(key? : string) {
```
```suggestion
        BigInt.prototype.toJSON = function(key?: string) {
```

src/modules/data/Height.ts
```diff
@@ -64,4 +64,15 @@ export class Height
     {
         return Utils.UInt64ToString(this.value);
     }
+
+    /**
+     * Converts this object to its JSON representation
```
TrustHenry, line 69:
```suggestion
     * Converts this object to it's JSON representation
```

src/modules/data/Height.ts
```diff
@@ -64,4 +64,15 @@ export class Height
...
+     *
+     * Use `string` as primitive types, as JS is only precise up to
+     * `2 ** 53 - 1` but we can get numbers up to `2 ** 64 - 1`.
+     */
+    public toJSON (key?: string) : string
```
TrustHenry:
```suggestion
    public toJSON (key?: string): string
```

TrustHenry (approved):
Looks good, just some styling nits.

src/modules/data/Height.ts
Geod24: In this case, `its` is correct.
https://www.grammarly.com/blog/its-vs-its
And yes, it's (!) weird to me too ;)

TrustHenry: Oh, TIL

Geod24:
Comments addressed

PR #150, opened 2020-10-28
ValidatorData: Use `Height` instead of bare UInt64

Geod24:


TrustHenry (approved):
Only ValidatorData looks good to use Height.
`UInt64` is still used a lot in our code instead of `Height`.

Geod24:
Yep, I have changes locally to completely remove `UInt64`, but I need to fix some tests, so I am doing the changes incrementally.

PR #153, opened 2020-10-28
Simplify a Storage test by using chained Promises

Geod24:
```
Currently, Stoa uses an odd mix of Promises and callback,
that carries the worst of both world with it.
Very little Promise chaining is done, as most usage wraps Promises.
Entangling this is quite perilous, but an easy entry point is the test suite,
as Mocha already supports returning Promises.
```

@MichaelKim20 @TrustHenry : https://stackoverflow.com/questions/52641469/how-to-test-async-code-with-mocha-using-await

### DIFF
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -37,52 +37,37 @@ describe ('Test ledger storage and inquiry function.', () =>
         ledger_storage.close();
     });
 
-    it ('Test for saving of all blocks', (doneIt: () => void) =>
+    it ('Test for saving of all blocks', () =>
     {
-        (() => {
-            return new Promise<void>(async (resolve, reject) =>
+        return new Promise<void>(async (resolve, reject) =>
+        {
+            try
             {
-                try
+                for (let elem of sample_data_raw)
                 {
-                    for (let elem of sample_data_raw)
-                    {
-                        let sample_data_item = JSON.parse(elem.replace(/([\[:])?(\d+)([,\}\]])/g, "$1\"$2\"$3"));
-                        await ledger_storage.putBlocks(sample_data_item);
-                    }
+                    let sample_data_item = JSON.parse(elem.replace(/([\[:])?(\d+)([,\}\]])/g, "$1\"$2\"$3"));
+                    await ledger_storage.putBlocks(sample_data_item);
                 }
-                catch (error)
+            }
+            catch (error)
+            {
+                reject(error);
+                return;
+            }
+            resolve();
+        }).then(() => {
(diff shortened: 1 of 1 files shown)


tests/Storage.test.ts
```diff
@@ -37,52 +37,37 @@ describe ('Test ledger storage and inquiry function.', () =>
         ledger_storage.close();
     });
 
-    it ('Test for saving of all blocks', (doneIt: () => void) =>
+    it ('Test for saving of all blocks', () =>
```
MichaelKim20, line 40: If the execution of this function is delayed and not completely terminated, then an error will occur.
The reason why 'doneIt()' is inserted is to call 'doneIt()' when the test is completed.
This ensures that the following tests can be run reliably
In fact, in the code below, problems occurred in the following tests.
``` TypeScript

        return new Promise<void>( (resolve, reject) =>
        {
            setTimeout(async() => {
                try {
                    for (let elem of sample_data_raw) {
                        let sample_data_item = JSON.parse(elem.replace(/([\[:])?(\d+)([,\}\]])/g, "$1\"$2\"$3"));
                        await ledger_storage.putBlocks(sample_data_item);
                    }
                }
                catch (error) {
                    reject(error);
                    return;
                }
                resolve();
            }, 2000);
        }).then(() => {
            let height_value = 1;
            let height = new Height(BigInt(height_value));
            ledger_storage.getBlocks(height)
                .then((rows: any[]) =>
                {
                    assert.strictEqual(rows.length, 1);
                    assert.strictEqual(rows[0].height, height_value);
                    assert.strictEqual(Hash.createFromBinary(rows[0].merkle_root, Endian.Little).toString(),
                        '0x9c4a20550ac796274f64e93872466ebb551ba2cd3f2f051533d07a478d2402b' +
                        '59e5b0f0a2a14e818b88007ec61d4a82dc9128851f43799d6c1dc0609fca1537d');
                })
        });
```

Geod24: Mocha will wait for `Promise`s if `Promise`s are returned to it. It's in the link I posted with the PR.

Geod24:
I raised this PR because I started to see race condition when removing some `sleep` calls in the test-suite.
So I started to look into the source and this was the first improvement needed. The next one will be to set different port per test.

tests/Storage.test.ts
MichaelKim20: You're right, I tested it in the wrong way,
I forgot the timeout was 2,000 ms.

MichaelKim20 (approved):

PR #154, opened 2020-10-28
Make 'WebService.start' Promise asynchronous

Geod24:
This allows us to integrate it better with Mocha and to delay
the error messages to the main in dev / prod,
and to let Mocha handle it (without exiting the process) in tests.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -85,10 +85,8 @@ class Stoa extends WebService
 
     /**
      * Setup and start the server
-     *
-     * @param callback An optional callback to register as listener
      */
-    public start (callback?: Function)
+    public async start (): Promise<void>
     {
         // Prepare middleware
 
@@ -106,7 +104,7 @@ class Stoa extends WebService
         this.app.post("/preimage_received", this.putPreImage.bind(this));
 
         // Start the server
-        super.start(callback);
+        return super.start();
     }
 
     /**
--- a/src/modules/service/WebService.ts
+++ b/src/modules/service/WebService.ts
@@ -58,40 +58,20 @@ export class WebService
     }
 
     /**
-     * Start web server
-     *
-     * @param callback An optional callback to register as listener
+     * Asynchronously start the web server
      */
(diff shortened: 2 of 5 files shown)


src/modules/service/WebService.ts
```diff
@@ -58,40 +58,20 @@ export class WebService
     }
 
     /**
-     * Start web server
-     *
-     * @param callback An optional callback to register as listener
+     * Asynchronously start the web server
      */
-    public start (callback?: Function)
+    public async start ()
```
MichaelKim20:
```suggestion
    public async start (): Promise<void>
```

src/modules/service/WebService.ts
```diff
@@ -58,40 +58,20 @@ export class WebService
...
-            }
-        });
-
         // Listen on provided this.port on this.address.
-        this.server.listen(this.port, this.address, () =>
-        {
-            logger.info(`Listening to requests on: ${this.address}:${this.port}`);
-            if (callback !== undefined)
-                callback();
+        return new Promise((resolve, reject) => {
```
MichaelKim20:
```suggestion
        return new Promise<void>((resolve, reject) => {
```

src/Stoa.ts
```diff
@@ -88,7 +88,7 @@ class Stoa extends WebService
      *
      * @param callback An optional callback to register as listener
      */
-    public start (callback?: Function)
+    public async start ()
```
MichaelKim20:
```suggestion
    public async start (): Promise<void>
```

src/Stoa.ts
```diff
@@ -88,7 +88,7 @@ class Stoa extends WebService
      *
      * @param callback An optional callback to register as listener
```
MichaelKim20:
```suggestion
```

MichaelKim20 (approved):
It's very good. It's good to be simple. I think there are some codes to change.

Geod24:
Comments addressed

PR #155, opened 2020-10-28
Modify the default configFile

TrustHenry:


src/modules/common/Config.ts
```diff
@@ -80,7 +80,7 @@ export class Config implements IConfig
         // Parse the arguments
         const parser = new ArgumentParser();
         parser.add_argument('-c', '--config', {
-            default: "config.yaml",
+            default: "docs/config.example.yaml",
```
Geod24, line 83: No no no! That's not a valid default.

TrustHenry: Well, do you have any other good ideas?

Geod24: Leave it as is.

Geod24: What's the motivation behind this PR ?

TrustHenry: It's not important.
It is because of the convenience of starting npm from the user's point of view.
This PR can be closed.

PR #157, opened 2020-10-28
Simplify Storage test setup

Geod24:
```
There is no need to chain Promises of async of promises.
A single `async` function with the right `await` in a callback
will do the job just fine.
```

Note, the 3 `catch` clauses can be removed because anything that is `thrown` will generate an implicit error message if not handled, and that can only happen if `getEnrollments` fails.


TrustHenry (approved):
Simple pretty!

MichaelKim20 (approved):

PR #158, opened 2020-10-28
 Remove TaskManager and rely on Promises instead

Geod24:
There is no need for such a construct in Javascript,
as the runtime provides such a mechanism, unlike in D.

The `TaskManager` seems to originate from Agora's similar concept,
which is simply a mean to abstract away the underlying event loop,
be it Vibe.d or Agora.

In Javascript, asynchronicity is part of the language,
e.g. `window.setInterval` / `window.setTimeout`,
hence we don't swap implementation like Agora would.
Instead of re-building an event-loop within JS,
just use the native constructs.


src/Stoa.ts
```diff
@@ -448,11 +445,9 @@ class Stoa extends WebService
...
                     {
                         // Recovery is required for blocks that are not received.
-                        let success: boolean = await this.recoverBlock(block, height, expected_height);
-                        if (!success)
-                        {
-                            this.pool.unshift(stored_data);
-                        }
+                        let success: boolean = false;
+                        while (!success)
+                            success = await this.recoverBlock(block, height, expected_height);
```
Geod24, line 450: I'm not 100% sure of this retry logic though. It matches the previous retry logic, but we should re-evaluate this in the future.

src/Stoa.ts
```diff
@@ -32,14 +31,14 @@ class Stoa extends WebService
...
     /**
-     * The temporary storage of the data received from Agora
-     */
-    protected pool: Array<IPooledData>;
-
-    /**
-     * The task manager that periodically executes registered a task function
+     * Chain of pending store operations
+     *
+     * To ensure switft response time to Agora when our handlers are called,
```
AndrejMitrovic:
```suggestion
     * To ensure swift response time to Agora when our handlers are called,
```

AndrejMitrovic (changes requested):
Makes sense to me. Just a typo to fix.

Geod24:
Fixed typo, and extended commit description / title.

Issue #137, opened 2020-10-23
No feedback when Stoa cannot connect to Agora


Geod24:
If Stoa starts and is unable to connect to Agora, this should be apparent to the user.
I would favor simply terminating the program after logging a proper error message.
In the future we can revisit this behavior (e.g. output messages and try reconnecting), but the current situation where nothing happens nor is displayed is very misleading.

Geod24:
Fixed by #161

Issue #140, opened 2020-10-23
Stoa never actually recovers block without push


Geod24:
The function supposed to do this, `recoverBlock`, is called only within `task`. However, Stoa will never actively query an Agora node, so between its start and the next push by Agora, Stoa won't be able to server any data.

PR #161, opened 2020-10-29
Stoa: Ensure we can connect to Agora on startup

Geod24:
This is a collection of commit to fix issue #137 .
This test for Agora being online when Stoa starts, and paves the way for #140 
Note that in a perfect world, Stoa wouldn't need Agora to work, but currently the error handling is quite missing, so it's better to have it this way for the time being, and make it more flexible later.


Geod24:
Self-merging as I want to test it live

PR #163, opened 2020-10-30
Change the parameter of `LedgerStorage.putBlocks` to an instance of the `Block`

MichaelKim20:
The method LedgerStorage.putBlocks has one parameter. This has data from the block.
Previously this was the JSON Object. I changed this to an instance of Block.
In order to deliver the correct type and create a more stable Stoa, I revised it like this.

### DIFF
--- a/tests/Storage.test.ts
+++ b/tests/Storage.test.ts
@@ -41,7 +41,7 @@ describe ('Test ledger storage and inquiry function.', () =>
                 for (let elem of sample_data_raw)
                 {
                     let sample_data_item = JSON.parse(elem.replace(/([\[:])?(\d+)([,\}\]])/g, "$1\"$2\"$3"));
-                    await ledger_storage.putBlocks(sample_data_item);
+                    await ledger_storage.putBlocks(Block.fromJSON(sample_data_item));
                 }
             }
             catch (error)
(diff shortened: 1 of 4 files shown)


tests/Storage.test.ts
```diff
@@ -245,21 +245,21 @@ describe ('Test for storing block data in the database', () =>
...
                 message: "SQLITE_CONSTRAINT: UNIQUE constraint failed: blocks.height"
             });
     });
 
     it ('DB transaction test when writing a block', async () =>
     {
-        let block = new Block();
         let sample_data0 = JSON.parse(sample_data_raw[0].replace(/([\[:])?(\d+)([,\}\]])/g, "$1\"$2\"$3"));
-        block.parseJSON(sample_data0);
+        let block = Block.fromJSON(sample_data0);
```
Geod24:
```suggestion
        const block = Block.fromJSON(sample_data0);
```

tests/Storage.test.ts
```diff
@@ -245,21 +245,21 @@ describe ('Test for storing block data in the database', () =>
     it ('Error-handling test when writing a block.', async () =>
     {
         let sample_data0 = JSON.parse(sample_data_raw[0].replace(/([\[:])?(\d+)([,\}\]])/g, "$1\"$2\"$3"));
-        await ledger_storage.putBlocks(sample_data0);
-        await assert.rejects(ledger_storage.putBlocks(sample_data0),
+        let block = Block.fromJSON(sample_data0);
```
Geod24, line 248:
```suggestion
        const block = Block.fromJSON(sample_data0);
```

Geod24 (approved):
Nice ! Just two comments. I was exploring doing something similar but couldn't get it to work.

Geod24:
Hold on, I'm going to do the fixes here directly and merge, as we need this deployed soon.

TrustHenry:
The comments have not been applied yet.

Geod24:
Pushed, will merge when green

tests/Storage.test.ts
TrustHenry, line 248: I think it's missing. ;)

Geod24: Oops, I actually caught a third instance :D

PR #166, opened 2020-10-30
Replace all usages of UInt64 with BigInt

Geod24:
We can greatly simplify the code by using the builtin Javascript type (at the expense of dropping compatibility with older Node.js versions).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2835,11 +2835,6 @@
       "resolved": "https://registry.npmjs.org/sprintf-js/-/sprintf-js-1.0.3.tgz",
       "integrity": "sha1-BOaSb2YolTVPPdAVIDYzuFcpfiw="
     },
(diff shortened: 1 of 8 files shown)


TrustHenry:
https://github.com/bpfkorea/stoa/blob/ff8fc0fd1b33926c1584145ac0816db9a5101f80/src/Stoa.ts#L294
https://github.com/bpfkorea/stoa/blob/ff8fc0fd1b33926c1584145ac0816db9a5101f80/src/modules/agora/AgoraClient.ts#L99
Is it okay to leave this?

Geod24:
> Is it okay to leave this?

It shouldn't have any effect anymore. I will raise a follow-up PR to remove those, and use builtin JSON deserialization.

TrustHenry (approved):
LGTM

PR #170, opened 2020-11-03
Simplify `string` to Public Key / Signature / Hash conversion

Geod24:
See each commit for detail.

### DIFF
--- a/src/modules/data/Hash.ts
+++ b/src/modules/data/Hash.ts
@@ -56,29 +56,21 @@ export class Hash
     /**
      * Reads from the hex string
      * @param hex The hex string
-     * @param endian The byte order
      * @returns The instance of Hash
      */
-    public fromString (hex: string, endian?: Endian): Hash
+    public fromString (hex: string): Hash
     {
-        if (endian === undefined)
-            endian = Endian.Little;
-
-        readFromString(hex, this.data, endian);
+        readFromString(hex, this.data);
         return this;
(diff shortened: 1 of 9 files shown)


src/modules/data/Hash.ts
```diff
@@ -123,9 +115,9 @@ export class Hash
      * @param endian The byte order
```
TrustHenry: Need to erase this phrase.

TrustHenry: Need to remove this phrase.

src/modules/data/PublicKey.ts
```diff
@@ -37,17 +37,32 @@ export class PublicKey
     public static Width: number = sodium.crypto_sign_PUBLICKEYBYTES;
 
     /**
-     * Constructor
-     * @param bin The binary data of the public key
+     * Construct a new instance of this class
+     *
+     * @param data   - The string or binary representation of the public key
+     * @param endian - The byte order
```
TrustHenry: JSDoc comment also supports `-`.
Reading is good if `-` are included.
However, most documents do not have `-`.
https://jsdoc.app/tags-param.html

Geod24: Indeed, will revert

Geod24:
Comments addressed.

TrustHenry (approved):
LGTM

PR #171, opened 2020-11-03
Improve json parsing of PreImageInfo

Geod24:
Currently, we allocate a dummy object and parse into it. The right way to do it in JS is through the `reviver` parameter to `JSON.parse`. Since the data in the test-suite is not actually a JSON string but actually an untyped object, the first step is to call reviver manually, and later on, once we switch to string, we can use `JSON.parse`.

### DIFF
--- a/src/modules/data/PreImageInfo.ts
+++ b/src/modules/data/PreImageInfo.ts
@@ -44,18 +44,23 @@ export class PreImageInfo
     }
 
     /**
-     * This parses JSON.
-     * @param json The object of the JSON
-     * @returns The instance of PreImageInfo
+     * The reviver parameter to give to `JSON.parse`
+     *
+     * This function allows to perform any necessary conversion,
+     * as well as validation of the final object.
+     *
+     * @param key   Name of the field being parsed
+     * @param value The value associated with `key`
+     * @returns A new instance of `PreImageInfo` if `key == ""`, `value` otherwise.
      */
-    public parseJSON (json: any): PreImageInfo
+    public static reviver (key: string, value: any): any
     {
-        Validator.isValidOtherwiseThrow<IPreImageInfo>('PreImageInfo', json);
+        if (key !== "")
+            return value;
 
-        this.enroll_key.fromString(json.enroll_key);
-        this.hash.fromString(json.hash);
-        this.distance = Number(json.distance);
+        Validator.isValidOtherwiseThrow<IPreImageInfo>('PreImageInfo', value);
 
-        return this;
+        return new PreImageInfo(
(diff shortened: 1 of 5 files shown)


src/modules/data/PreImageInfo.ts
```diff
@@ -21,41 +21,53 @@ import { Validator, IPreImageInfo } from './validator'
...
+     *
+     * @param key   Name of the field being parsed
+     * @param value The value associated with `key`
+     * @returns A new instance of `PreImageInfo` if `key == ""`, `value` otherwise.
      */
-    public parseJSON (json: any): PreImageInfo
+    public static reviver (key: string, value: any): any
     {
-        Validator.isValidOtherwiseThrow<IPreImageInfo>('PreImageInfo', json);
+        if (key !== "")
```
TrustHenry, line 65: Do you allow whitespaces in the empty string check method?

src/modules/data/PreImageInfo.ts
```diff
@@ -21,41 +21,53 @@ import { Validator, IPreImageInfo } from './validator'
...
-     * @param json The object of the JSON
-     * @returns The instance of PreImageInfo
+     * The reviver parameter to give to `JSON.parse`
+     *
+     * This function allows to perform any necessary conversion,
+     * as well as validation of the final object.
+     *
+     * @param key   Name of the field being parsed
+     * @param value The value associated with `key`
+     * @returns A new instance of `PreImageInfo` if `key == ""`, `value` otherwise.
```
TrustHenry, line 61: A Document description is the looks like a code. :)

src/modules/data/PreImageInfo.ts
Geod24: No, there's no reason to do so.

src/modules/data/PreImageInfo.ts
```diff
@@ -21,41 +21,53 @@ import { Validator, IPreImageInfo } from './validator'
...
      */
-    public parseJSON (json: any): PreImageInfo
+    public static reviver (key: string, value: any): any
```
TrustHenry, line 63: Key value is not being used.

Geod24: ???
It is used

Geod24:
In case there's some confusion, here's the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

src/modules/data/PreImageInfo.ts
TrustHenry: Ah~ yes, Using the reviver parameter

TrustHenry (approved):
LGTM.
I did understand the of this PR codes.

PR #172, opened 2020-11-03
Use the reviver approach everywhere and make ctor parameters mandatory

Geod24:
See each commit for exanation. Follow what was started / experimented in #171


TrustHenry (approved):
Nice improvement, Thanks!

PR #156, opened 2020-10-28
fix .gitignore

TrustHenry:

PR #176, opened 2020-11-04
Move test data in their own folder/file

Geod24:
This makes the TS source code more readable by separating those giant strings.


.gitignore
```diff
@@ -113,4 +113,4 @@ database
 src/modules/data/types
 
 # Database files
-data/
+/data/
```
TrustHenry, line 116: Isn't it a folder?

Geod24: It means everything under this path. `data/` match any data folder in the repo (including `tests/data/`) while `/data/` only match `data/` at the root of the repository.

TrustHenry: Yes,
But in my case the root data folder at root `/data/` was not ignored by git.

Geod24: Could you demonstrate how it doesn't work for you ?

TrustHenry: https://github.com/bpfkorea/stoa/pull/156
There used to be `stoa/logs/` in front of.
I think it was a different case.

tests/Utils.ts
```diff
@@ -12,13 +12,15 @@
...
+export const sample_data_raw = (() => {
+    return [
+        fs.readFileSync('tests/data/Block.0.sample1.json', 'utf-8'),
```
TrustHenry, line 20: Nice!

TrustHenry (approved):

Issue #165, opened 2020-10-30
Sending blocks to Stoa failed on live system


Geod24:
```
Oct 30 03:56:53 eu-002 docker[9687]: [Stoa] 2020-10-30T03:56:53.221Z info Connected to Agora, block height is 0
Oct 30 03:56:53 eu-002 docker[9687]: [Stoa] 2020-10-30T03:56:53.223Z info Listening to requests on: 0.0.0.0:2828
Oct 30 04:00:49 eu-002 docker[9687]: [Stoa] 2020-10-30T04:00:49.975Z info Recovered a block with block height of 0
Oct 30 04:00:49 eu-002 docker[9687]: (node:38) UnhandledPromiseRejectionWarning: Error: SQLITE_ERROR: cannot rollback - no transaction is active
Oct 30 04:00:49 eu-002 docker[9687]: (Use `node --trace-warnings ...` to show where the warning was created)
Oct 30 04:00:49 eu-002 docker[9687]: (node:38) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
Oct 30 04:00:49 eu-002 docker[9687]: (node:38) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Looks like a serialization error which might be solved once #163 is merged.

TrustHenry:
While the agora block height is 3 and stoa is performing the recover block operation, there seems to be an error while saving the block of 0 height.

Geod24:
Now the error message is better:
```
Oct 30 07:28:03 eu-002 docker[8956]: (node:38) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 6)
Oct 30 07:28:03 eu-002 docker[8956]: (node:38) UnhandledPromiseRejectionWarning: Error: Validation failed: TxInputs - should be string
Oct 30 07:28:03 eu-002 docker[8956]:     at Function.isValidOtherwiseThrow (/stoa/bin/src/modules/data/validator/index.ts:95:23)
[...]

PR #178, opened 2020-11-09
Use the received data without replacing it

MichaelKim20:
We received the data as a string, replaced the number with a string, and then did JSON.parse.
But this part is not necessary because UInt64 is no longer used. So I removed this.

The reason for the error in the issue is that the attribute value was numbered only when it was written in letters.
I think this is a problem caused by receiving mixed data in various forms. This will also be resolved by removing the obvious substitution of this part.

Relates to #165

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -86,8 +86,9 @@ class Stoa extends WebService
         // Prepare middleware
 
         // parse application/x-www-form-urlencoded
-        this.app.use(bodyParser.urlencoded({extended: false}));
-        this.app.use(bodyParser.raw({type: "*/*"}));
+        this.app.use(bodyParser.urlencoded({ extended: false }))
+        // parse application/json
+        this.app.use(bodyParser.json())
         this.app.use(cors(cors_options));
         // enable pre-flight
         this.app.options('*', cors(cors_options));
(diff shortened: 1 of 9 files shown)


MichaelKim20:
I was able to get the results of the test as follows.

```
[Stoa] 2020-11-09T02:16:29.846Z http POST /blocks_externalized block=[object Object]
[Stoa] 2020-11-09T02:16:29.906Z info Recovered a block with block height of 0
[Stoa] 2020-11-09T02:16:29.912Z info Recovered a block with block height of 1
[Stoa] 2020-11-09T02:16:29.918Z info Recovered a block with block height of 2
[Stoa] 2020-11-09T02:16:29.922Z info Saved a block with block height of 3
[Stoa] 2020-11-09T02:17:25.693Z http POST /blocks_externalized block=[object Object]
[Stoa] 2020-11-09T02:17:25.704Z info Saved a block with block height of 4
```

MichaelKim20:
I only ran Agora, and created blocks up to height 2.
I executed Agora and Stoa, and created heigth 3 block.
0, 1, and 2 were recovered and 3 were recorded.

src/Stoa.ts
```diff
@@ -86,8 +86,10 @@ class Stoa extends WebService
         // Prepare middleware
 
         // parse application/x-www-form-urlencoded
-        this.app.use(bodyParser.urlencoded({extended: false}));
-        this.app.use(bodyParser.raw({type: "*/*"}));
+        // parse application/x-www-form-urlencoded
```
Geod24: Duplicated comment

Geod24 (approved):
Nice! Just a minor comment.

TrustHenry (approved):
LVGTM!

src/Stoa.ts
MichaelKim20: I fixed it

Issue #168, opened 2020-11-02
Adapt Stoa to changes to `Input`


Geod24:
Stoa relied on "txhash + index" scheme, so the changes for for bpfk/agora#1286 broke it.

PR #179, opened 2020-11-09
Adapt Stoa to changes to Agora's `Input`

TrustHenry:
The Struct change of the block caused a change in the data of the block required for the test.
The block data used in the test was get from TestNet.
This PR is necessary to move forward to the next stage of improvement.
Stoa does not rely on the index of the input.

Related to #168

### DIFF
--- a/tests/Stoa.test.ts
+++ b/tests/Stoa.test.ts
@@ -77,9 +77,9 @@ describe ('Test of Stoa API Server', () =>
         client.get (uri.toString())
             .then((response) =>
             {
-                assert.strictEqual(response.data.length, 3);
+                assert.strictEqual(response.data.length, 6);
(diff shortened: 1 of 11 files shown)


tests/Stoa.test.ts
```diff
@@ -187,45 +187,54 @@ describe ('Test of Stoa API Server', () =>
...
+            doneIt();
+        }, 200);
+
+        /**
+         * To do
+         * The preimage_received service was interrupted briefly.
+         * See Stoa.ts putPreImage(req, res);
+         */
+
+        /**
```
Geod24: ?

TrustHenry: The annotation modified.

MichaelKim20:
LGTM

PR #185, opened 2020-11-11
Reset start height in an incomplete recovery

MichaelKim20:
If the number of blocks to be recovered is too large only a part of them will be recovered.
Therefore, the height of the block to start the recovery is taken from the database.

Currently, I have found a problem that does not stop when the number of blocks to be restored is greater than 64(Stoa._max_count_on_recovery).


TrustHenry (approved):
LGTM

PR #186, opened 2020-11-11
Change error status code for Get request

### LINKED ISSUE #139: No error code returned on wrong height
Currently, starting a Stoa without Agora, and doing a `curl`, a.k.a:
```shell
$ curl http://127.0.0.1:4242/validators
```

I get an empty response on the query side, but can see that the request was processed:
```
[Stoa] 2020-10-23T04:15:10.214Z http GET /validators
```

Returning no data and no error code is surprising to the user, which is likely to always expect something.
While a bare `/validators` call should always succeed (currently fails because of #137), the following might not:
```
$ curl http://127.0.0.1:4242/validators?height=4242424242
```
And should definitely return an error code. This obviously apply to all endpoints.

TrustHenry:
The GET /validators call should always be successful,
but Stoa cannot be serviced if Stoa is listening to the block or if there is a problem with the DB.
The GET /validator may request an incorrect height.
In this case, return the 400 status code and respond that there is no data.

Fixed #139

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -133,8 +133,7 @@ class Stoa extends WebService
         if ((req.query.height !== undefined) &&
             !Utils.isPositiveInteger(req.query.height.toString()))
         {
-            res.status(400).send({
-                statusMessage: `Invalid value for parameter 'height': ${req.query.height.toString()}`});
+            res.status(400).send(`Invalid value for parameter 'height': ${req.query.height.toString()}`);
             return;
         }
 
@@ -152,7 +151,11 @@ class Stoa extends WebService
                 // Nothing found
                 if (!rows.length)
                 {
-                    res.status(204).send();
+                    if (height !== null)
+                        res.status(400).send("No validator exists for block height.");
+                    else
+                        res.status(503).send("Stoa is currently unavailable.");
+
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


src/Stoa.ts
```diff
@@ -232,7 +231,7 @@ class Stoa extends WebService
                 // Nothing to show
                 if (!rows.length)
                 {
-                    res.status(204).send();
+                    res.status(400).send("The validator data not found.");
```
MichaelKim20: How about adding `address`?

TrustHenry: I don't know what it means. Give me a suggestion.

src/Stoa.ts
```diff
@@ -145,7 +144,8 @@ class Stoa extends WebService
                 // Nothing found
                 if (!rows.length)
                 {
-                    res.status(204).send();
+                    // Service Unavailable
+                    res.status(503).send("Stoa is currently unavailable.");
```
MichaelKim20: How about this phrase?
`No validator exists for block height.`

MichaelKim20: It would be nice if the block height was also displayed.

src/Stoa.ts
MichaelKim20: `The validator data not found. (${address})`

MichaelKim20: I think this would be better.
`The validator data not found. (address: ${address}, height: ${height})`

src/Stoa.ts
TrustHenry: Block height is optional for parameters.
Also, the `/validators` service should always be serviceable. 
The lack of data can mean that the Stoa is preparing or the DB is empty.

src/Stoa.ts
TrustHenry: The requester will of course know the value of the parameter.
I don't think it's a bad idea to mark it.

src/Stoa.ts
TrustHenry: However, this may require a response to an incorrect request height.

MichaelKim20 (approved):
LGTM

PR #188, opened 2020-11-12
[WIP] Remove the catch-up function from the test code

MichaelKim20:
I checked late after seeing intermittent errors in some test codes when github's CI was executed due to the catch-up function.

I was able to remove the catch-up function from the test code to solve this problem.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -116,7 +116,7 @@ class Stoa extends WebService
             .then(
                 () =>
                 {
-                    return this.pending = this.pending.then(() => { return this.catchup(height); });
+                    return this.catchup(height);
                 });
     }
 
--- a/tests/Stoa.test.ts
+++ b/tests/Stoa.test.ts
@@ -35,13 +35,17 @@ describe ('Test of Stoa API Server', () =>
     let agora_server: TestAgora;
     let client = axios.create();
 
-    before ('Start Stoa API Server and a fake Agora', () =>
+    before ('Start a fake Agora', () =>
     {
-        let prom = new Promise<void>((resolve, reject) => {
+        return new Promise<void>((resolve, reject) => {
             agora_server = new TestAgora("2826", sample_data, resolve);
         });
+    });
+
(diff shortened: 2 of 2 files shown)


Geod24:
If it fails in the CI, it's probably because there's a bug. Avoiding the call will not solve it, just hide it.

MichaelKim20:
This is a problem caused by Mocha waiting only for the first Promise in the process of processing Promise chain.
So I plan to make an issue of this.

MichaelKim20:
```TypeScript
describe ('Test of Promise in Mocha', () =>
{
    before("begin", () => {
        function p1():Promise<void>
        {
                setTimeout(() => {
                    console.log('Promise 1');
                    resolve();
                }, 100);
            })
        }

        function p2():Promise<void>
        {
                setTimeout(() => {
                    console.log('Promise 2');
                    resolve();
                }, 200);
            });
        }

        function p3(): Promise<void>
        {
                setTimeout(() => {
                    console.log('Promise 3');
                    resolve();
                }, 300);
            });
        }

        return p1()
            .then(() => { return p2; })
            .then(() => { return p3; });
    });

    it ('it 2', () =>
    {
        console.log("it 2");
    })
});

describe ('Test of Promise in Mocha2', () =>
{
    it ('it 1', () =>
    {
        function p1():Promise<void>
        {
                setTimeout(() => {
                    console.log('Promise 1');
                    resolve();
                }, 100);
            })
        }
        function p2():Promise<void>
        {
                setTimeout(() => {
                    console.log('Promise 2');
                    resolve();
                }, 200);
            });
        }

        function p3(): Promise<void>
        {
                setTimeout(() => {
                    console.log('Promise 3');
                    resolve();
                }, 300);
            });
        }


        return p1()
            .then(() => { return p2; })
            .then(() => { return p3; });
    });

    it ('it 2', () =>
    {
        console.log("it 2");
    })
});
```
Result is 
Promise 1
it 2

MichaelKim20:
Since this is new, I made it proceed in issue #190

Issue #183, opened 2020-11-10
The current block height of the Stoa is required.


TrustHenry:
`/block_height`

It is necessary to provide `height` information for the received blocks of the Stoa.

Geod24:
That's a good point. Let's add it to the next sprint.

PR #193, opened 2020-11-18
Implement block_height API endpoint

TrustHenry:
It is necessary to provide height information for the received blocks of the Stoa.

Related to #183

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -94,6 +94,7 @@ class Stoa extends WebService
         this.app.options('*', cors(cors_options));
 
         // Prepare routes
+        this.app.get("/block_height", this.getBlockHeight.bind(this));
         this.app.get("/validators", this.getValidators.bind(this));
         this.app.get("/validator/:address", this.getValidator.bind(this));
         this.app.post("/block_externalized", this.putBlock.bind(this));
@@ -359,6 +360,29 @@ class Stoa extends WebService
 
         res.status(200).send();
     }
+
+    /**
+     * GET /block_height
+     *
+     * Return the highest block height stored in Stoa
+     */
+    private getBlockHeight (req: express.Request, res: express.Response)
+    {
+        logger.http(`GET /block_height`);
+
+        this.ledger_storage.getBlockHeight()
+            .then((row: Height | null) => {
+                if (row == null)
+                    res.status(400).send(`The block height not found.`);
+                else
+                    res.status(200).send(JSON.stringify(row));
+            })
+            .catch((err) => {
+                logger.error("Failed to data lookup to the DB: " + err);
+                res.status(500).send("Failed to data lookup");
+            }
+            );
+    };
 
     /**
      * Extract the block height from JSON.
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -219,6 +219,32 @@ export class LedgerStorage extends Storages
     }
 
     /**
+     * Get the highest block height in this Stoa DB
+     * @returns Returns the Promise. If it is finished successfully the `.then`
+     * of the returned Promise is called with the block height
+     * and if an error occurs the `.catch` is called with an error.
+     */
+    public getBlockHeight (): Promise<Height | null>
+    {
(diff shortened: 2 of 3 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -218,6 +218,29 @@ export class LedgerStorage extends Storages
...
+                .then((row: any[]) =>
+                {
+                    resolve(new Height(BigInt(row[0].height)));
+                })
+                .catch((err) =>
+                {
+                    reject(err);
+                });
+        });
+    }
```
MichaelKim20: If the data does not exist, the value of 'height' will be null.
If the value of 'height' is 'null', an error occurs when 'null' is converted to 'BigInt'.
Then, go to the 'catch' below and call reject(err).
Therefore, if there is no data, 'http status 400' is not delivered and '500' is delivered.
How about the following suggestion?
```suggestion
    public getBlockHeight (): Promise<Height | null>
    {
        return new Promise<Height | null>((resolve, reject) =>
        {
            let sql = `SELECT MAX(height) as height FROM blocks`;
            this.query(sql, [])
                .then((row: any[]) =>
                {
                    if (row[0].height !== null)
                        resolve(new Height(BigInt(row[0].height)));
                    else
                        resolve(null);
                })
                .catch((err) =>
                {
                    reject(err);
                });
        });
    }
```

src/Stoa.ts
```diff
@@ -360,6 +361,31 @@ class Stoa extends WebService
...
+        logger.http(`GET /block_height`);
+
+        this.ledger_storage.getBlockHeight()
+            .then((row: any) => {
+                if (!row)
+                {
+                    res.status(400).send(`The block data not found.`);
+                    return;
+                }
+                res.status(200).send(JSON.stringify(row));
```
MichaelKim20: Following my suggestion for LedgerStorage.getBlockHeight, I suggest:

```suggestion
            .then((row: Height | null) => {
                if (row == null)
                    res.status(400).send(`The block data not found.`);
                else
                    res.status(200).send(JSON.stringify(row));
```

src/Stoa.ts
```diff
@@ -360,6 +361,31 @@ class Stoa extends WebService
...
     }
 
+    /**
+     * GET /block_height
+     *
+     * Return the highest block height stored in Stoa
+     */
+    private getBlockHeight (req: express.Request, res: express.Response)
+    {
+        logger.http(`GET /block_height`);
```
MichaelKim20, line 371: Good to leave a log.

src/modules/storage/LedgerStorage.ts
```diff
@@ -218,6 +218,29 @@ export class LedgerStorage extends Storages
...
+     * Get the highest block height in this Stoa DB
+     * @returns Returns the Promise. If it is finished successfully the `.then`
+     * of the returned Promise is called with the block height
+     * and if an error occurs the `.catch` is called with an error.
+     */
+    public getBlockHeight (): Promise<Height>
+    {
+        return new Promise<Height>((resolve, reject) =>
+        {
+            let sql = `SELECT MAX(height) as height FROM blocks`;
```
MichaelKim20, line 231: How about the query below?
`SELECT value FROM information WHERE key = 'height';`

tests/Stoa.test.ts
```diff
@@ -67,6 +67,25 @@ describe ('Test of Stoa API Server', () =>
         });
     });
 
+    it ('Test of the path /block_height', (doneIt: () => void) =>
+    {
+        let uri = URI(host)
+            .port(port)
+            .directory("block_height");
```
MichaelKim20: There is also a method called `filename` in URI, isn't this appropriate?

MichaelKim20 (commented):
Overall, it's a good structure. I have a few suggestions.

TrustHenry:
@MichaelKim20 Thanks for reviews. Comments addressed

src/modules/storage/LedgerStorage.ts
TrustHenry: That's good, but I want to check the actual stored data.

MichaelKim20 (approved):
LGTM

Issue #180, opened 2020-11-09
Merge common code (data structures) between SDK and Stoa


Geod24:
Seems like Stoa and the SDK have a lot of duplication, essentially:
- https://github.com/bpfkorea/boa-sdk-ts/tree/v0.x.x/src/modules/data
- https://github.com/bpfkorea/stoa/tree/v0.x.x/src/modules/data

Looks like we could easily have Stoa depends on part of the SDK and reduce code duplication.

PR #197, opened 2020-11-23
Adapt the file KeyPair.ts of BOA-SDK-TS to Stoa

MichaelKim20:
I adapted the file `KeyPair.ts` of BOA-SDK-TS to Stoa.
I replaced with `libsodium-wrappers-sumo`, the same sodium library used by BOA-SDK-TS
Because there is no `crypto_sign_ed25519_sk_to_seed` in the existing sodium library, it is replaced with a library that supports it.

Relates to #180


MichaelKim20:
I changed the library and made a docker image. It was confirmed that it was produced quickly normally.
And block data received from Agora was being processed normally.

MichaelKim20:
I rebased this PR

src/main.ts
```diff
@@ -35,24 +36,30 @@ const stoa: Stoa = new Stoa(config.database.filename,
     config.server.address
     );
 
-stoa.createStorage().then(() => {
-    stoa.start().then(() => {
-        logger.info(`Listening to requests on: ${config.server.address}:${config.server.port}`);
+SodiumHelper.init().then(() => {
+    stoa.createStorage().then(() => {
```
Geod24, line 40: Does `createStorage` depends on Sodium working ?

Geod24:
Why is `KeyPair` needed in Stoa ?
Also, instead of copying over files, we should includes the SDK from Stoa, so we don't duplicate code.

MichaelKim20:
> Why is `KeyPair` needed in Stoa ?
> Also, instead of copying over files, we should includes the SDK from Stoa, so we don't duplicate code.

What do you think about replacing Stoa's sodium library with `libsodium-wrappers-sumo`?

MichaelKim20:
@Geod24: Here's my plan.
I will first make a repository named `boa-common-ts`.
I will copy the data module and the utility module here.
I will register this in the npm module.
I have a plan to install and use these npm modules in Stoa and SDK.
What do you think of this plan? If you agree with this plan, please review the following as well.
The hindrance to this is the sodium library. But if we use `libsodium-wrappers-sumo`, this will also be solved.
I would appreciate it if you could proceed with the review of `libsodium-wrappers-sumo`.

Geod24:
Why do we need a third package ? Cannot Stoa just depends on the SDK ?

Geod24:
> I would appreciate it if you could proceed with the review of libsodium-wrappers-sumo.

Yes this part is fine 👍

src/main.ts
MichaelKim20: Sodium is WebAssembly.
It takes some waiting time for it to load into the machine.
Although the `createStorage` is not dependent on sodium, the hash function is then used when blocks are received from the catch-up function.
I think it would be better to put the sodium work in the front rather than add it in the middle.

MichaelKim20:
I don't think this PR needs to be reviewed anymore because there is no need to add a common npm module of the STOA and SDK. I'll close this.

PR #198, opened 2020-11-23
Refactory data modules

MichaelKim20:
I worked on the ability to easily create a Transaction object and also convert it to JSON.
I added the methods `toJSON` for JSON serialization.
I renamed TxOutputs to TxOutput, TxInputs to TxInput.
I added a method that returns an instance with initialized values.
I enabled the constructor of `TxInput` to be overload.

Relates to #180

### DIFF
--- a/src/modules/data/Transaction.ts
+++ b/src/modules/data/Transaction.ts
@@ -11,8 +11,8 @@
 
 *******************************************************************************/
 
-import { TxInputs } from './TxInputs';
-import { TxOutputs } from './TxOutputs';
+import { TxInput } from './TxInput';
+import { TxOutput } from './TxOutput';
 import { DataPayload } from './DataPayload';
 import { Validator, ITransaction } from './validator'
 
@@ -42,12 +42,12 @@ export class Transaction
     /**
      * The array of references to the unspent output of the previous transaction
      */
-    public inputs: TxInputs[];
+    public inputs: TxInput[];
 
     /**
      * The array of newly created outputs
      */
-    public outputs: TxOutputs[];
+    public outputs: TxOutput[];
 
     /**
      * The data payload to store
@@ -61,7 +61,7 @@ export class Transaction
      * @param outputs - The array of newly created outputs
      * @param payload - The data payload to store
      */
-    constructor (type: number, inputs: TxInputs[], outputs: TxOutputs[], payload: DataPayload)
+    constructor (type: number, inputs: TxInput[], outputs: TxOutput[], payload: DataPayload)
     {
         this.type = type;
         this.inputs = inputs;
@@ -88,8 +88,8 @@ export class Transaction
 
         return new Transaction(
             Number(value.type),
-            value.inputs.map((elem: any) => TxInputs.reviver("", elem)),
-            value.outputs.map((elem: any) => TxOutputs.reviver("", elem)),
+            value.inputs.map((elem: any) => TxInput.reviver("", elem)),
+            value.outputs.map((elem: any) => TxOutput.reviver("", elem)),
             DataPayload.reviver("", value.payload));
     }
 
--- a/src/modules/data/Transaction.ts
+++ b/src/modules/data/Transaction.ts
@@ -106,4 +106,17 @@ export class Transaction
             elem.computeHash(buffer);
         this.payload.computeHash(buffer);
     }
+
+    /**
+     * Converts this object to its JSON representation
+     */
+    public toJSON (key?: string): object
+    {
+        return {
+            "type": this.type,
+            "inputs": this.inputs,
+            "outputs": this.outputs,
+            "payload": this.payload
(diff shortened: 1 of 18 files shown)


src/modules/data/Transaction.ts
```diff
@@ -42,12 +42,12 @@ export class Transaction
     /**
      * The array of references to the unspent output of the previous transaction
      */
-    public inputs: TxInputs[];
+    public inputs: TxInput[];
```
TrustHenry, line 45:
```suggestion
    public inputs: Input[];
```
https://github.com/bpfkorea/agora/blob/cb259d978d5e9240a46be2a309c5e08370477cd9/source/agora/consensus/data/Transaction.d#L52

Is there a reason why you don't use the same name as Agora?

src/modules/storage/LedgerStorage.ts
```diff
@@ -491,7 +491,7 @@ export class LedgerStorage extends Storages
         }
 
         function save_input (storage: LedgerStorage, height: Height, tx_idx: number,
-            input: TxInputs, out_index: number): Promise<void>
+                             input: TxInput, out_index: number): Promise<void>
```
```suggestion
        input: TxInput, out_index: number): Promise<void>
```

src/modules/data/Transaction.ts
MichaelKim20: It will avoid colliding with other packages of the symbol.

src/modules/storage/LedgerStorage.ts
MichaelKim20: I fixed it

src/modules/data/TxInput.ts
```diff
@@ -36,13 +37,36 @@ export class TxInputs
 
     /**
      * Constructor
-     * @param utxo - The hash of the UTXO to be spent
-     * @param signature - A signature that should be verified using public key of the output in the previous transaction
+     * @param first  UTXO or hash of transaction or instance of Transaction
+     * @param second Instance of Signature or index of Outputs
```
TrustHenry, line 41: It needs a more detailed document.

TrustHenry:
Rebase is required.

src/modules/data/TxInput.ts
MichaelKim20: I added a more detailed document.

MichaelKim20:
I rebased this PR.

TrustHenry (approved):
LGTM

src/modules/data/TxInput.ts
```diff
@@ -0,0 +1,129 @@
...
+            this.signature = second;
+        }
+        // constructor (first: Transaction, second: bigint) {}
+        // first is the instance of the previous transaction
+        // second is the index of the output in the previous transaction
+        else if ((first instanceof Transaction) && (typeof second === 'bigint'))
+        {
+            this.utxo = makeUTXOKey(hashFull(first), second);
+            this.signature = Signature.init;
+        }
```
Geod24, line 60: IMO this jungle of `if` on types is a big red flag. It completely nullify the benefit of having a typed language, which is to catch errors at compile time, not run time.

If we want to write code that uses an overload-like approach, okay, but it should be done using the language feature that allows it: https://www.typescriptlang.org/docs/handbook/functions.html#overloads

src/modules/data/TxOutput.ts
```diff
@@ -22,7 +21,7 @@ import { SmartBuffer } from 'smart-buffer';
  * Convert JSON object to TypeScript's instance.
  * An exception occurs if the required property is not present.
  */
-export class TxOutputs
+export class TxOutput
```
Geod24, line 24: 👍

Geod24:
Please, do not write TypeScript code like we write D code. Those are two very different languages and we can't just use the same idioms in both. Not to say some idioms can't be used, but I see a lot of code being needlessly complicated because it's attempting to be written the same way as it is in D.

Also, why do we need the `toJSON` methods ? It looks like in at least some places, the method that was added does exactly the same thing as the default behavior.

MichaelKim20:
> Please, do not write TypeScript code like we write D code. Those are two very different languages and we can't just use the same idioms in both. Not to say some idioms can't be used, but I see a lot of code being needlessly complicated because it's attempting to be written the same way as it is in D.
> 
> Also, why do we need the `toJSON` methods ? It looks like in at least some places, the method that was added does exactly the same thing as the default behavior.

The reason for adding this is to easily change the object to JSON string. In the case of Transaction, it was difficult with a single method call because of the mix of array and object. And there's a test code for this.
https://github.com/bpfkorea/stoa/blob/v0.x.x/tests/ValidationJSON.test.ts#L154-L179

Geod24:
> The reason for adding this is to easily change the object to JSON string.

We can have `toJSON` hooks, but there is no reason to add them if they just do what the default is.

Issue #190, opened 2020-11-16
Correct failures that occur in test code


MichaelKim20:
In the process of processing PR #188, 
I confirmed that when I used the Promise chain in Mocha, I did not wait for all Promise to end.
We will have to find another alternative to this part.

Geod24:
Can you past a test code ? Mocha works fine with Promises, I never saw a bug. So I think the usage of Promise might have been incorrect instead.

PR #200, opened 2020-11-23
Fix failures that occur in test code

MichaelKim20:
1. I changed to separate fake Agora and Stoa when preparing for the test.
    This is to avoid using the Promise-chain in Mocha.
2. I detached a method for creating LedgerStorage from the Stoa constructor
    This is to proceed with follow-up work after the creation of LedgerStorage is completed. 
    Because Promise cannot be returned from the constructor of Stoa, a separate method was implemented.

Relates to #190

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -22,7 +22,7 @@ declare global {
 
 class Stoa extends WebService
 {
-    public ledger_storage: LedgerStorage;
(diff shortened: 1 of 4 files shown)


MichaelKim20:
I've run 10 times on Github's CI.
I checked that there was no error at all.

TrustHenry (approved):
Nice

PR #202, opened 2020-11-23
Separate types associated with Stoa API responses from common types

MichaelKim20:
I separated the response data of Stoa service from the data module.
Because the purpose of this is one of the tasks required to package common data


TrustHenry (approved):

Geod24:
Nice!

PR #208, opened 2020-11-27
Logging only when a new pre-image is updated in the database

TrustHenry:
The problem was that It used fat arrow for callback declaration
Using the otherwise anonymous function, this is bounded in dynamic scope and 
so `this.changes` is valued.

If it updated the pre-image, return the updated count.
This corrects the error of printing the log that has always been saved.

Related to https://github.com/bpfkorea/agora/issues/1322

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -16,6 +16,7 @@ import {
     TxInput, TxOutput, makeUTXOKey, hashFull, TxType,
     Utils, Endian
 } from 'boa-sdk-ts';
+import { logger } from '../common/Logger';
 import { Storages } from './Storages';
 
 /**
@@ -340,10 +341,10 @@ export class LedgerStorage extends Storages
     /**
      * Update a preImage to database
      */
-    public updatePreImage (pre_image: PreImageInfo): Promise<void>
+    public updatePreImage (pre_image: PreImageInfo): Promise<number>
     {
         let enroll_key = pre_image.enroll_key.toBinary(Endian.Little);
-        return new Promise<void>((resolve, reject) =>
+        return new Promise<number>((resolve, reject) =>
         {
             this.run(
                 `UPDATE validators
@@ -374,9 +375,9 @@ export class LedgerStorage extends Storages
                     pre_image.distance,
                     pre_image.distance
                 ])
-                .then(() =>
+                .then((result) =>
                 {
-                    resolve();
+                    resolve(result.changes);
                 })
                 .catch((err) =>
                 {
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -597,9 +597,10 @@ class Stoa extends WebService
                 try
                 {
                     let pre_image = PreImageInfo.reviver("", stored_data.data);
+                    let changes = await this.ledger_storage.updatePreImage(pre_image);
 
-                    await this.ledger_storage.updatePreImage(pre_image);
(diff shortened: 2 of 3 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -374,8 +375,11 @@ export class LedgerStorage extends Storages
                     pre_image.distance,
                     pre_image.distance
                 ])
-                .then(() =>
+                .then((result) =>
                 {
+                    if (result.changes == 1)
+                        logger.info(`Saved a pre-image enroll_key : ${pre_image.enroll_key.toString().substr(0, 18)}, ` +
+                        `hash : ${pre_image.hash.toString().substr(0, 18)}, distance : ${pre_image.distance}`);
```
MichaelKim20: I would like to be able to set the permissions for log output in Stoa

TrustHenry: Yes, that would be good.
I applied it now.

TrustHenry:
@Geod24 @MichaelKim20 Ready for review.
And Check the issue https://github.com/bpfkorea/agora/issues/1322

MichaelKim20:
How about changing the title of PR and Commit to the next phrase?
`Logging only when a new pre-image is updated in the database`

MichaelKim20:
The first and second commit is not related to the issue, so it should be handled as a separate PR.

TrustHenry:
> The first and second commit is not related to the issue, so it should be handled as a separate PR.

I think there is no problem as each commit contains documentation.

MichaelKim20:
> The problem was that It used fat arrow for callback declaration
> Using the otherwise anonymous function, this is bounded in dynamic scope and
> so `this.changes` is valued.
> 
> Related to [bpfkorea/agora#1322](https://github.com/bpfkorea/agora/issues/1322)

There is no explanation for the cause and resolution of this issue. There is only an explanation for the additional things.

TrustHenry:
> > The problem was that It used fat arrow for callback declaration
> > Using the otherwise anonymous function, this is bounded in dynamic scope and
> > so `this.changes` is valued.
> > Related to [bpfkorea/agora#1322](https://github.com/bpfkorea/agora/issues/1322)
> 
> There is no explanation for the cause and resolution of this issue. There is only an explanation for the additional things.

@MichaelKim20 Thanks for reviews.
Added document to target commit.

MichaelKim20:
Isn't it an issue that Agora frequently sends pre-images?
And isn't the solution you chose only to print the log when it's updated in the database?
I don't think the bug is the direct cause of the issue.

TrustHenry:
> Isn't it an issue that Agora frequently sends pre-images?
> 
> And isn't the solution you chose only to print the log when it's updated in the database?
> 
> I don't think the bug is the direct cause of the issue.

https://github.com/bpfkorea/agora/issues/1322#issuecomment-734656382

MichaelKim20:
It's a good result that you found and solved a problem that is not easy.
But the cause of the issue does not seem to be the problem.

The method you have adopted to resolve the issue seems to be to log only when the pre-image is updated in the database and the number of changed records is greater than 0.

And the bug where it returns the number of changed records has been corrected. This is not a problem that caused the issue, but I think you have found and dealt with things that are not easy. I applaud your achievements.

TrustHenry:
The style fix commit was moved.

TrustHenry:
> But the cause of the issue does not seem to be the problem.

Yes, the problem should I solved continuously in Agora.

MichaelKim20 (approved):

Geod24 (approved):

PR #209, opened 2020-11-28
Add code preparing sodium library to test code

MichaelKim20:
I added the preparatory work that was missing from the test code.


Geod24 (approved):
LGTM, although it looks like user code might be negatively affected.

PR #213, opened 2020-12-08
Fix bug where catching up does not work in special case

MichaelKim20:
Special case:
Agora only has Genesis blocks, and Stoa doesn't have any.


TrustHenry (approved):
Good.
There has been a problem with catch-up depending on when Agora and Stoa are running.

TrustHenry:
No related issues. Can you make a simple one?

MichaelKim20:
> No related issues. Can you make a simple one?

Thank you.
I didn't make an issue because I found a bug and could solve it in such a short time.

Issue #212, opened 2020-12-07
Implement shell function to provide transactions details of specific address


MichaelKim20:
For web wallet development, we need an API endpoint that returns the transactions for a given PublicKey.

Example:

``` TypeScript
class Tx {
    hash : Hash	                    // The hash of the transaction
    height: bigint                  // The height of the block
    time: Date                      // The time of block created
    type: number                    // The type of transaction
    source : PublicKey              // Source address
    destination : PublicKey         // Destination address
    amount: bigint                  // Amount
    fee: bigint                     // Fee
}
```

Input parameters : page-size, page-index, and address.
The return type : Tx[]

We're going to use shell methods, for the time being, the way we handled them at https://github.com/bpfkorea/boa-sdk-ts/issues/16.

let base = new Date("2020-01-01 00:00:00");
let height = 1;
let time = new Date(base.getTime() + height x 10 x 60000);

AndrejMitrovic:
`class Txs` I think you mean `class Tx`. This is the return value for just one transaction, correct?

AndrejMitrovic:
How about the `type`? It can be a pay or a freeze.

MichaelKim20:
@AndrejMitrovic: A transaction has multiple inputs.
Therefore, we do not know which input address paid the fee.
Therefore, we cannot show the fees for individual transfers, but we should show the fee for the transaction on the block explorer site.
Therefore, it would be better to remove the fee from the above class.
What do you think?

AndrejMitrovic:
> Therefore, it would be better to remove the fee from the above class.

It would be much easier if Tx was the real Transaction with Inputs and Outputs + an additonal timestamp field. Then you would simply return all transactions which have at least one Input or Output that references that key. 

The wallet can easily calculate the fee, this saves network bandwidth which is important. Your own solution creates many `Tx` with duplicate hashes.

The wallet itself can define how to display that data.

MichaelKim20:
I studied this issue more and reviewed the UI of the wallet.
As a result, we found that the following output was necessary.


**Data structure for the transactions history**

```TypeScript
/**
 * The interface of the transactions history element
 */
export interface ITxHistoryElement
{
    /**
     * Address, Public key
     */
    address: string;

    /**
     * Block height
     */
    height: string;

    /**
     * Transaction time
     */
    time: number;

    /**
     * Transaction hash
     */
    tx_hash: string;

    /**
     * Transaction type
     */
    type: number;

    /**
     * Amount
     */
[...]

PR #214, opened 2020-12-08
Add an endpoint for providing the transaction history and overview

MichaelKim20:
I saw the plan of the wallet and realized it to be as supportive as possible.

Relates to #212

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -4,7 +4,7 @@ import { LedgerStorage } from './modules/storage/LedgerStorage';
 import { logger } from './modules/common/Logger';
 import { Height, PreImageInfo, Hash, hash, Block, Utils, Endian } from 'boa-sdk-ts';
 import { WebService } from './modules/service/WebService';
-import { ValidatorData, IPreimage, IUnspentTxOutput } from './Types';
+import { ValidatorData, IPreimage, IUnspentTxOutput, ITxHistoryElement } from './Types';
 
 import bodyParser from 'body-parser';
 import cors from 'cors';
@@ -123,6 +123,7 @@ class Stoa extends WebService
         this.app.get("/validators", this.getValidators.bind(this));
         this.app.get("/validator/:address", this.getValidator.bind(this));
         this.app.get("/utxo/:address", this.getUTXO.bind(this));
+        this.app.get("/wallet/transactions/history/:addresses", this.getWalletTransactionsHistory.bind(this));
         this.app.post("/block_externalized", this.putBlock.bind(this));
         this.app.post("/preimage_received", this.putPreImage.bind(this));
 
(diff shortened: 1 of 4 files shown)


MichaelKim20:
I added two endpoints.

### 1. GET /wallet/transactions/history/:addresses
An endpoint that provides a history of multiple transactions

- The parameter `addresses` are the form of multiple addresses separated by `,`.
- The parameter `pageSize` is the maximum size that can be obtained from one query, default is 10
- The parameter `page` is the number on the page, this value begins with 1, default is 1
- The return type is array of ITxHistoryElement.

```TypeScript
/**
 * The interface of the transactions history element
 */
export interface ITxHistoryElement
{
    /**
     * Address, Public key
     */
    address: string;

    /**
     * Block height
     */
    height: string;

    /**
     * Transaction time
     */
    time: number;

    /**
     * Transaction hash
     */
    tx_hash: string;

    /**
     * Transaction type
     */
    type: number;

    /**
     * Amount
     */
    amount: string;

    /**
     * Block height at which the output of the transaction becomes available
     */
    unlock_height: string;

    /**
     * Time at which the output of the transaction becomes available
     */
    unlock_time: number;
}
```
---
### 2. GET /wallet/transaction/overview/:hash
An endpoint that provides an overview of one transaction
- The parameter `hash` is the hash of the transaction
- The return type is ITxOverview.
``` TypeScript
/**
 * The interface of the transaction overview
 */
export interface ITxOverview
{
    /**
     * Block height
     */
    height: string;

    /**
     * Transaction time
     */
    time: number;

    /**
     * Transaction hash
     */
    tx_hash: string;

    /**
     * Transaction type
     */
    type: number;

    /**
     * Block height at which the output of the transaction becomes available
     */
    unlock_height: string;

    /**
     * Time at which the output of the transaction becomes available
     */
    unlock_time: number;

    /**
     * The address and amount of the output associated with the transaction input
     */
    senders: Array<ITxOverviewElement>;

    /**
     * The address and amount of transaction output
     */
    receivers: Array<ITxOverviewElement>;

    /**
     * Transaction fee
     */
    fee: string;
}

/**
 * The interface of the transaction overview element
 */
export interface ITxOverviewElement
{
    /**
     * Address, Public key
     */
    address: string;

    /**
     * Amount
     */
    amount: string;
}
```

MichaelKim20:
The file below is the database I used to test the query.

[main.db.zip](https://github.com/bpfkorea/stoa/files/5664202/main.db.zip)
[transactions-history5.sql.zip](https://github.com/bpfkorea/stoa/files/5664223/transactions-history5.sql.zip)

src/modules/storage/LedgerStorage.ts
```diff
@@ -849,4 +849,137 @@ export class LedgerStorage extends Storages
...
+                        SUM(S.amount) as spend
+                    FROM
+                        transactions AS T,
+                        tx_inputs AS I,
+                        tx_outputs AS S
+                    WHERE
+                        T.block_height = I.block_height
+                        AND T.tx_index = I.tx_index
+                        AND I.utxo = S.utxo_key
+                        AND S.address in (${accounts})
```
TrustHenry: You must also join the blocks table.
It's not on the block table, but there may only be transactions, tx_input, tx_ouputs

TrustHenry: Is it the standard JOIN method for sqlite?
Shouldn't we do this?
FROM transactions T INNER JOIN tx_inputs I ON (T.block_height = I.block_height)

MichaelKim20: > You must also join the blocks table.
> It's not on the block table, but there may only be transactions, tx_input, tx_ouputs

When is that the case?

MichaelKim20: > Is it the standard JOIN method for sqlite?
> Shouldn't we do this?
> FROM transactions T INNER JOIN tx_inputs I ON (T.block_height = I.block_height)

That's a good idea.

TrustHenry: For data integrity.

src/modules/storage/LedgerStorage.ts
```diff
@@ -849,4 +849,137 @@ export class LedgerStorage extends Storages
...
+            (
+                    SELECT
+                        S.address,
+                        T.block_height,
+                        T.tx_hash,
+                        T.tx_index,
+                        T.type,
+                        T.unlock_height,
+                        0 as income,
+                        SUM(S.amount) as spend
```
```suggestion
                        IFNULL(SUM(S.amount),0) as spend
```

src/modules/storage/LedgerStorage.ts
MichaelKim20: Okay, That's a good idea

src/modules/storage/LedgerStorage.ts
```diff
@@ -849,4 +849,137 @@ export class LedgerStorage extends Storages
...
+                    UNION ALL
+
+                    SELECT
+                        O.address,
+                        T.block_height,
+                        T.tx_hash,
+                        T.tx_index,
+                        T.type,
+                        T.unlock_height,
+                        SUM(O.amount) as income,
```
```suggestion
                        IFNULL(SUM(O.amount),0) as income,
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -849,4 +849,137 @@ export class LedgerStorage extends Storages
...
+                        tx_outputs AS O
+                    WHERE
+                        T.block_height = O.block_height
+                        AND T.tx_index = O.tx_index
+                        AND O.address in (${accounts})
+                    GROUP BY T.tx_hash, O.address
+            ) AS TX
+            GROUP BY TX.block_height, TX.tx_hash, TX.address, TX.type
+            ORDER BY TX.block_height DESC, TX.tx_index DESC
+            limit ? offset ?;`;
```
```suggestion
            LIMIT ? OFFSET ?;`;
```

MichaelKim20: @TrustHenry Thank you for review
I changed

MichaelKim20:
I changed the layout of the table for the integrity of the data
And it exists at the beginning of the commit.

docs/Database_Schema.md
```diff
@@ -67,9 +67,10 @@ CREATE TABLE IF NOT EXISTS "enrollments" (
 
 | Column            | Data Type | PK | Not NULL | Default  |Description|
 |:----------------- |:--------- |:--:|:--------:| -------- | --------- |
-| block_height      | INTEGER   | Y  | Y        |          | The height of the block|
-| tx_index          | INTEGER   | Y  | Y        |          | The index of transaction in the block |
-| tx_hash           | BLOB      |    | Y        |          | The hash of transaction |
+| tx_hash           | BLOB      | Y  | Y        |          | The hash of transaction |
+| tx_index          | INTEGER   |    | Y        |          | The index of transaction in the block |
+| block_hash        | BLOB      |    | Y        |          | The hash of the current block |
```
TrustHenry: This column is unnecessary from the db normalization point of view.
The primary key of the block is still defined by `height`.
If you want to use this as an FK, you must change the PK in the block table to Block_hash.

MichaelKim20: That's a good point.
If I modify this, the changes are too big, so I will create a new issue and proceed.
I will remove the commit from the current PR.

src/Stoa.ts
```diff
@@ -377,6 +379,143 @@ class Stoa extends WebService
...
+     * GET /wallet/transaction/history/:addresses
+     *
+     * Called when a request is received through the `/transaction_history/:addresses` handler
+     * The parameter `addresses` are the form of multiple addresses separated by `,`.
+     * The parameter `pageSize` is the maximum size that can be obtained from one query, default is 10
+     * The parameter `page` is the number on the page, this value begins with 1, default is 1
+     *
+     * Returns a set of transactions history of the addresses.
+     */
+    private getWalletTransactionHistory (req: express.Request, res: express.Response)
```
```suggestion
    private getWalletTransactionsHistory (req: express.Request, res: express.Response)
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -849,4 +849,142 @@ export class LedgerStorage extends Storages
...
+            `SELECT
+                T.block_height as height,
+                T.tx_hash,
+                T.type,
+                T.unlock_height
+            FROM
+                (
+                    SELECT * FROM transactions WHERE tx_hash = ?
+                ) T
+                INNER JOIN blocks B ON (B.height = T.block_height);`;
```
```suggestion
            FROM 
                blocks B INNER JOIN transactions T
                ON (B.height = T.block_height)
            WHERE T.tx_hash= ? ;`;
```

MichaelKim20: Why?

TrustHenry (approved):
LGTM

PR #217, opened 2020-12-15
Add a field to the results of an endpoint

MichaelKim20:
GET /wallet/transactions/history/:addresses
Three additional fields have been added:

```
    // The transaction type of wallet ('inbound', 'outbound', 'freeze', 'payload')
    wallet_tx_type: string
    // The address that sent (or received) the funds
    peer: string;
    // The number of peers
    peer_count: number;
```

GET /wallet/transaction/overview/:hash
The following field has been added
```
    // The hash of UTXO
    utxo: string;
    // The transaction data payload
    payload: string;

```

PR #220, opened 2020-12-16
Add filtering capabilities for wallet endpoint `/wallet/transactions/history/:address` 

MichaelKim20:
Based on wallet requirements, it implements filtering capabilities for endpoint `/wallet/transactions/history/:addresses` implemented for wallets.

### From `Wallet Requirements Documentation` 
**Filters**
● The ability to filter by **​Transaction type**​. The list show be additive, e.g. one can select both inbound and
outbound transaction, excluding freeze and data;
● A **date range**, comprising of a start date and an end date;
● The ability to filter by a single **peer**, which can be specified by name, or by address. Partial addresses (a
substring of the address) must be supported;

**A detailed description of the field**
● Transaction type​: A visual representation (through a colored icon) of the type of transaction. If the transaction is inbound (funds are added), a green plus sign (“+”) is used. For outbound funds, a red minus sign (“-“) is used. For freezing transactions, a blue snowflake icon should be used. For data storage, the icon and color are not yet defined.
● Peer​: The address that sent (or received) the funds. If there is an associated name, the name must be
displayed.

Dependent on #217

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -124,7 +124,7 @@ class Stoa extends WebService
         this.app.get("/validators", this.getValidators.bind(this));
         this.app.get("/validator/:address", this.getValidator.bind(this));
         this.app.get("/utxo/:address", this.getUTXO.bind(this));
-        this.app.get("/wallet/transactions/history/:addresses", this.getWalletTransactionsHistory.bind(this));
+        this.app.get("/wallet/transactions/history/:address", this.getWalletTransactionsHistory.bind(this));
         this.app.get("/wallet/transaction/overview/:hash", this.getWalletTransactionOverview.bind(this));
         this.app.post("/block_externalized", this.putBlock.bind(this));
         this.app.post("/preimage_received", this.putPreImage.bind(this));
@@ -381,11 +381,11 @@ class Stoa extends WebService
     }
 
     /**
-     * GET /wallet/transactions/history/:addresses
+     * GET /wallet/transactions/history/:address
      *
-     * Called when a request is received through the `/wallet/transactions/history/:addresses` handler
+     * Called when a request is received through the `/wallet/transactions/history/:address` handler
      * ```
-     * The parameter `addresses` are the form of multiple addresses separated by `,`.
+     * The parameter `address` are the address to query.
      * The parameter `pageSize` is the maximum size that can be obtained
      *      from one query, default is 10
      * The parameter `page` is the number on the page, this value begins with 1,
@@ -409,9 +409,9 @@ class Stoa extends WebService
      */
     private getWalletTransactionsHistory (req: express.Request, res: express.Response)
     {
-        let params_addresses: string = String(req.params.addresses);
+        let address: string = String(req.params.address);
 
-        logger.http(`GET /wallet/transactions/history/${params_addresses}}`);
+        logger.http(`GET /wallet/transactions/history/${address}}`);
 
         let filter_begin: number | undefined;
         let filter_end: number | undefined;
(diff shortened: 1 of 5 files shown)


MichaelKim20:
I changed the parameter address to support only one from several.
I found that I had to do this in order to provide a clear results.

src/Stoa.ts
```diff
@@ -381,59 +381,109 @@ class Stoa extends WebService
...
+        let address: string = String(req.params.address);
 
-        logger.http(`GET /wallet/transactions/history/${params_addresses}}`);
+        logger.http(`GET /wallet/transactions/history/${address}}`);
 
-        let page_size = (req.query.pageSize !== undefined)
-            ? Number(req.query.pageSize.toString())
-            : 10;
+        if ((req.query.beginDate !== undefined) &&
+            !Utils.isPositiveInteger(req.query.beginDate.toString()))
```
TrustHenry: Can it verify the correct date integer format? Validates only integer formats.
An error must occur if the start date is greater than the end date. or the opposite case
What if the Genesis Block is less than the start time?

src/Stoa.ts
```diff
@@ -381,59 +381,109 @@ class Stoa extends WebService
...
+        if ((req.query.endDate !== undefined) &&
+            !Utils.isPositiveInteger(req.query.endDate.toString()))
+        {
+            res.status(400).send(`Invalid value for parameter 'endDate': ${req.query.endDate.toString()}`);
+            return;
+        }
 
-        if (page_size <= 0)
+        if ((req.query.pageSize !== undefined) &&
+            !Utils.isPositiveInteger(req.query.pageSize.toString()))
```
TrustHenry: Don't you limit the maximum size?

src/Stoa.ts
```diff
@@ -381,59 +381,109 @@ class Stoa extends WebService
...
         {
-            res.status(400).send(`Positive pageSize expected: ${page_size}`);
+            res.status(400).send(`Invalid value for parameter 'pageSize': ${req.query.pageSize.toString()}`);
             return;
         }
 
-        if (page_size > 30)
+        if ((req.query.page !== undefined) &&
+            !Utils.isPositiveInteger(req.query.page.toString()))
```
TrustHenry: This is allowed if the page size is larger than the page size if entered.

src/Stoa.ts
```diff
@@ -381,59 +381,109 @@ class Stoa extends WebService
...
         {
-            res.status(400).send(`Page size cannot be a number greater than 30: ${page_size}`);
+            res.status(400).send(`Invalid value for parameter 'page': ${req.query.page.toString()}`);
             return;
         }
 
-        if (page <= 0)
+        let page_size = (req.query.pageSize !== undefined)
+            ? Number(req.query.pageSize.toString())
+            : 10;
```
TrustHenry: You can handle it together on line 430.

TrustHenry (commented):
a few comments.
I'll continue to review it.

src/Stoa.ts
MichaelKim20: When the two are combined, it becomes like this and the structure is not easy to understand, so we separated them.
And the top code is the verification code for when an error occurs when converted to a number.
I think it's better to separate the verification code from the actual code
``` TypeScript
        let page_size: number;
        if (req.query.pageSize !== undefined) {
            if (!Utils.isPositiveInteger(req.query.pageSize.toString())) {
                res.status(400).send(`Invalid value for parameter 'pageSize': ${req.query.pageSize.toString()}`);
                return;
            }
            page_size = Number(req.query.pageSize.toString())
        }
        else page_size = 10
```

src/Stoa.ts
MichaelKim20: I don't understand your question, so I'd appreciate it if you could write it more specifically.

TrustHenry: There is a typo. `page` cannot be larger than the `page size`.

src/Stoa.ts
MichaelKim20: >> What if the Genesis Block is less than the start time?

I don't think this will be a big problem. This is because there is no error during the data lookup process

TrustHenry: Yes, You can handle it later, but you need to check the validity of the date.

MichaelKim20: >> An error must occur if the start date is greater than the end date

That's a good suggestion. I'll apply it.

src/Stoa.ts
TrustHenry: There will now be more parameter verification routines.
I want to unify the validation routine.
But now we can go this way.

src/Stoa.ts
MichaelKim20: But you know this is different from other types, right?

MichaelKim20: But it's clear that suggestions are one of the best ways.

MichaelKim20:
I changed it all

TrustHenry (approved):
LGTM

PR #224, opened 2020-12-21
Create Stoa's Transactions Pool DB Schema Part1

TrustHenry:
This transaction pool stores transactions received from Agora.
It will be erased from the pool when it is later saved in the block.

### DIFF
--- a/docs/Database_Schema.md
+++ b/docs/Database_Schema.md
@@ -224,3 +224,78 @@ CREATE TABLE IF NOT EXISTS information (
     PRIMARY KEY(key)
 )
 ```
+
+----
+
+## 9. Table **transactionPool**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+| tx_hash           | BLOB      | Y  | Y        |          | The hash of transaction |
+| type              | INTEGER   |    | Y        |          | The type of transaction |
+| payload           | BLOB      |    | Y        |          | The transaction data payload |
+| time              | INTEGER   |    | Y        |          | Received time |
+
+### _Create Script_
+
+```sql
+CREATE TABLE IF NOT EXISTS "transactionPool" (
+    "tx_hash"               BLOB    NOT NULL,
+    "type"                  INTEGER NOT NULL,
+    "payload"               BLOB    NOT NULL,
+    "time"                  INTEGER NOT NULL,
+    PRIMARY KEY("tx_hash")
+)
+```
+
+----
+
+## 4. Table **txInputPool**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+| tx_hash           | BLOB      | Y  | Y        |          | The hash of transaction|
+| input_index       | INTEGER   | Y  | Y        |          | The index of input in the inputs|
+| utxo              | BLOB      |    | Y        |          | The hash of the UTXO to be spent|
+| signature         | BLOB      |    | Y        |          | The signature of this transaction input|
+### _Create Script_
+
+```sql
+CREATE TABLE IF NOT EXISTS "txInputPool" (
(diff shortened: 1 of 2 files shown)


docs/Database_Schema.md
```diff
@@ -224,3 +224,74 @@ CREATE TABLE IF NOT EXISTS information (
...
+----
+
+## 5. Table **txOutputPool**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+|  tx_hash          | BLOB      | Y  | Y        |          | The hash of transaction |
+|  output_index     | INTEGER   | Y  | Y        |          | The index of output in the outputs.|
```
MichaelKim20: Why do we need output_index?

docs/Database_Schema.md
```diff
@@ -224,3 +224,74 @@ CREATE TABLE IF NOT EXISTS information (
...
+----
+
+## 4. Table **txInputPool**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+| tx_hash           | BLOB      | Y  | Y        |          | The hash of transaction|
+| utxo              | BLOB      | Y  | Y        |          | The hash of the UTXO to be spent|
```
MichaelKim20: Why does this(utxo) have to be the key?

docs/Database_Schema.md
```diff
@@ -224,3 +224,74 @@ CREATE TABLE IF NOT EXISTS information (
...
+
+----
+
+## 9. Table **transactionPool**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+| tx_hash           | BLOB      | Y  | Y        |          | The hash of transaction |
```
MichaelKim20, line 236:
```suggestion
| time              | INTEGER   |    | Y        |          | Receive TimeStamp |
| tx_hash           | BLOB      | Y  | Y        |          | The hash of transaction |
```
Don't we need the time received?

MichaelKim20 (commented):
I have some questions.

docs/Database_Schema.md
TrustHenry: Yes, we need it.

docs/Database_Schema.md
TrustHenry: The outputs is an array.
A key is required to distinguish.

docs/Database_Schema.md
TrustHenry: Isn't it unique?

docs/Database_Schema.md
MichaelKim20: OKay

docs/Database_Schema.md
MichaelKim20: How about adding in_index just in case? I think we can print it out as JSON later.

TrustHenry: That's a good.

TrustHenry:
@MichaelKim20  Thanks for review. I apply the suggestions.

docs/Database_Schema.md
```diff
@@ -224,3 +224,78 @@ CREATE TABLE IF NOT EXISTS information (
...
+| signature         | BLOB      |    | Y        |          | The signature of this transaction input|
+### _Create Script_
+
+```sql
+CREATE TABLE IF NOT EXISTS "txInputPool" (
+    "tx_hash"               BLOB    NOT NULL,
+    "input_index"           INTEGER NOT NULL,
+    "utxo"                  BLOB    NOT NULL,
+    "signature"             BLOB    NOT NULL,
+    PRIMARY KEY("tx_hash","utxo")
```
MichaelKim20:
```suggestion
    PRIMARY KEY("tx_hash","input_index")
```
Isn't this what you want?

TrustHenry: Nice catch! Thank you.

MichaelKim20 (approved):

Issue #221, opened 2020-12-17
Implement /transaction_received endpoint


TrustHenry:
API to implement:

```
POST /transaction_received
```

Can Stoa create a transaction pool with it.
The transactions received are stored in the Stoa DB and can be cleared from the transaction pool when the blocks are externalized.
This allows the wallet to check the pending of the transaction.

PR #225, opened 2020-12-22
Store a transaction on the transaction pool

TrustHenry:
Store information, inputs, and outputs of received transactions on the tables.
Add endpoint /transaction_received

Related to #221

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -2,7 +2,7 @@ import { AgoraClient } from './modules/agora/AgoraClient';
 import { cors_options } from './cors';
 import { LedgerStorage } from './modules/storage/LedgerStorage';
 import { logger } from './modules/common/Logger';
-import { Height, PreImageInfo, Hash, hash, Block, Utils, Endian } from 'boa-sdk-ts';
+import { Height, PreImageInfo, Hash, hash, Block, Utils, Endian, Transaction, hashFull } from 'boa-sdk-ts';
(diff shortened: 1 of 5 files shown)


src/Stoa.ts
```diff
@@ -828,6 +850,23 @@ class Stoa extends WebService
                     reject(err);
                 }
             }
+            else if (stored_data.type === "transaction")
+            {
+                try
+                {
+                    let tx = Transaction.reviver("", stored_data.data);
+                    let changes = await this.ledger_storage.putTransactionPool(tx);
```
MichaelKim20, line 858: `changes ` is not used.

MichaelKim20 (commented):
Very good.
I think one test code should be added.
A test is required to verify that the database is stored normally.
And I have a question.
The same transaction will not be received more than once, right?

src/Stoa.ts
TrustHenry: Using the changes, I modified the code in many areas.
Validates the transaction inserted and the number of DB reflections of outputs & inputs

TrustHenry:
> Very good.
> I think one test code should be added.
> A test is required to verify that the database is stored normally.
> And I have a question.
> The same transaction will not be received more than once, right?

Yes, save the same transaction only once.
I will add a new test related to storage.

MichaelKim20 (approved):
LGTM

PR #229, opened 2020-12-24
Add DB Trigger for Transaction Pool

### LINKED ISSUE #227: When a block is received, The transactions removed from the transaction pool.
When a new block is received,
it is no longer needed because the pending transaction has been stored in the block.

It manages the transaction life cycle of the transaction pool.

TrustHenry:
A trigger occurs when a new block is received and a new transaction is inserted into the transaction table.
The trigger deletes the transactions in the transaction pool.
This knows whether a transaction in a transaction pool is in a pendding state.

Related to #227

### DIFF
--- a/docs/Database_Schema.md
+++ b/docs/Database_Schema.md
@@ -252,7 +252,7 @@ CREATE TABLE IF NOT EXISTS "transaction_pool" (
 
 ----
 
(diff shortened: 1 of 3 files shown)


MichaelKim20 (approved):
LGTM

Issue #231, opened 2020-12-29
Stoa cannot be restarted


Geod24:
Attempting to restart Stoa on the live system yields:
```
[Stoa] 2020-12-29T03:04:15.876Z info Agora endpoint: http://127.1.1.1:4567
[Stoa] 2020-12-29T03:04:15.877Z info sqlite3 database filename: /Users/geod24/projects/bpfk/stoa/data/main.db
[Stoa] 2020-12-29T03:04:15.947Z error Failed to create LedgerStorage: Error: SQLITE_ERROR: trigger tx_trigger already exists
```

This is blocking us from making any progress and should be fixed ASAP.

Geod24:
CC @MichaelKim20 @TrustHenry when will you have time to tackle this ?

TrustHenry:
I'll fix it.

Geod24:
Fixed in #233 although a test is still missing.

TrustHenry:
I will add a test case.

PR #233, opened 2020-12-29
 Do not create a trigger if it is already created.

TrustHenry:
A trigger cannot be created if a trigger exists.

Related to #231

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -167,7 +167,7 @@ export class LedgerStorage extends Storages
             PRIMARY KEY(tx_hash, output_index)
         );
 
-        CREATE TRIGGER tx_trigger AFTER INSERT ON transactions
+        CREATE TRIGGER IF NOT EXISTS tx_trigger AFTER INSERT ON transactions
         BEGIN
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
Why not just use `IF NOT EXISTS`? Wouldn't that work?

src/modules/storage/LedgerStorage.ts
```diff
@@ -167,6 +167,7 @@ export class LedgerStorage extends Storages
             PRIMARY KEY(tx_hash, output_index)
         );
 
+        DROP TRIGGER IF EXISTS tx_trigger;
         CREATE TRIGGER tx_trigger AFTER INSERT ON transactions
```
Geod24: Why do you drop it instead of using `IF NOT EXISTS` like the other tables ?

Geod24:
Also, tests

TrustHenry:
> Why not just use `IF NOT EXISTS`? Wouldn't that work?

Oh, I used the old method.
The latest version of sqlite is supported.

Geod24:
Also, commit title / PR title is not accurate anymore

MichaelKim20 (approved):

Geod24:
I still don't see a test that restart works. It's not the first time that a functionality as simple as restart gets broken, so please add a test.

PR #240, opened 2021-01-05
Change an endpoint `/utxo:address` considering pending transactions

### LINKED ISSUE #237: Do not reuse UTXO that has already been used in the wallet.
The UTXO used in pending transactions is still available to the endpoint `/utxo/:address`, which causes failure when sending a second transaction.

MichaelKim20:
I added fields (block height, time) to an endpoint `/utxo:address`
I added code to remove UTXO used in pending transactions.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -1019,19 +1019,24 @@ export class LedgerStorage extends Storages
      */
     public getUTXO (address: string): Promise<any[]>
     {
+        // TODO We should apply the block's timestamp to this method when it is added
         let sql =
             `SELECT
-                tx_outputs.utxo_key AS utxo,
-                transactions.type,
-                transactions.unlock_height,
-                tx_outputs.amount
+                O.utxo_key as utxo,
+                O.amount,
+                T.block_height,
+                STRFTIME('%s', '2020-01-01 00:00:00') * 1000 + T.block_height * 10 * 60000 as block_time,
+                T.type,
+                T.unlock_height
             FROM
-                tx_outputs,
-                transactions
+                tx_outputs O
+                INNER JOIN transactions T ON (T.tx_hash = O.tx_hash)
+                INNER JOIN blocks B ON (B.height = T.block_height)
             WHERE
-                tx_outputs.tx_hash = transactions.tx_hash
-                AND tx_outputs.address = ?
-                AND tx_outputs.used != 1;`;
+                O.address = ?
+                AND O.used = 0;
+            ORDER BY T.block_height, O.amount
+            `;
         return this.query(sql, [address]);
     }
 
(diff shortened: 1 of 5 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -1019,20 +1019,49 @@ export class LedgerStorage extends Storages
...
+            this.query(sql_utxo, [address])
+                .then((utxos: any[]) => {
+                    result = utxos;
+                    return this.query(sql_pending, [address]);
+                })
+                .then((pending: any[]) => {
+                    resolve(result.filter(n => pending.find(m => (Buffer.compare(n.utxo, m.utxo) === 0)) === undefined));
+                })
+                .catch(reject);
+        });
```
MichaelKim20, line 1063: Filtering was used because it was difficult to handle with one query statement.

src/modules/storage/LedgerStorage.ts
```diff
@@ -1019,20 +1019,48 @@ export class LedgerStorage extends Storages
...
+                tx_outputs O
+                INNER JOIN transactions T ON (T.tx_hash = O.tx_hash)
+                INNER JOIN blocks B ON (B.height = T.block_height)
             WHERE
-                tx_outputs.tx_hash = transactions.tx_hash
-                AND tx_outputs.address = ?
-                AND tx_outputs.used != 1;`;
-        return this.query(sql, [address]);
+                O.address = ?
+                AND O.used != 1;
```
```suggestion
                AND O.used = 0;
```

tests/Stoa.test.ts
```diff
@@ -393,3 +368,112 @@ describe ('Test of Stoa API Server', () =>
...
+                        type: 0,
+                        utxo: '0xd9482016835acc6defdfd060216a5890e00cf8f0a79ab0b83d3385fc723cd45bfea66eb3587a684518ff1756951d38bf4f07abda96dcdea1c160a4f83e377c32',
+                        amount: '24400000000000',
+                        height: '1',
+                        time: 1577837400000,
+                        unlock_height: '2',
+                        unlock_time: 1577838000000
+                    }
+                ]
+                assert.deepStrictEqual(response.data, expected);
```
TrustHenry: There is no unlock_time, but it is strange that the test code has passed.

src/modules/storage/LedgerStorage.ts
```diff
@@ -1019,20 +1019,48 @@ export class LedgerStorage extends Storages
...
+
+        let result: any[];
+        return new Promise<any[]>((resolve, reject) => {
+            this.query(sql_utxo, [address])
+                .then((utxos: any[]) => {
+                    result = utxos;
+                    return this.query(sql_pending, [address]);
+                })
+                .then((pending: any[]) => {
+                    resolve(result.filter(n => pending.find(m => (Buffer.compare(n.utxo, m.utxo) === 0)) === undefined));
```
TrustHenry, line 1060: What happens if `utxos` is null?

MichaelKim20: If the value is null, it means an error has occurred and it moves to catch.

MichaelKim20: So I don't think we need any more code.

TrustHenry (approved):
LGTM

PR #243, opened 2021-01-06
Change the value of the `lockfileVersion` to repair the docker build

MichaelKim20:
I changed lockfileVersion to 2


Geod24:
First, the CI needs to be setup to avoid having the same problem. Second, the issue seems to stem from the difference in versions (package.json was updated, but not package.lock). This has the same issue: package.json requires 0.0.11, but package.lock uses 0.0.10.

PR #210, opened 2020-11-28
Allow stoa to call getPreimage and improve TestAgora

Geod24:
Pretty straightforward, see each commit for details.

PR #245, opened 2021-01-06
Make getBlocksFrom return type typed

Geod24:
Since everyone needs to call Block.reviver otherwise.

Extracted from https://github.com/bpfkorea/stoa/pull/210


TrustHenry (approved):
LGTM

Issue #222, opened 2020-12-21
Apply to Stoa about Agora's flash layer


MichaelKim20:
Agora has developed a flash layer in a new branch. 
https://github.com/bpfkorea/agora/tree/flash-layer

Here it should be added to the store and implement the parts that need to be changed.

We will also need to work on BOA-SDK.

PR #246, opened 2021-01-07
Change to return `Promise` when testing asynchronous functions

MichaelKim20:
I propose a modified code for more accurate testing when testing asynchronous functions in test code.

### DIFF
--- a/tests/Stoa.test.ts
+++ b/tests/Stoa.test.ts
@@ -36,7 +36,7 @@ describe ('Test of Stoa API Server', () =>
     let agora_server: TestAgora;
     let client = axios.create();
 
-    before('Wait for the package libsodium to finish loading', () =>
+    before ('Wait for the package libsodium to finish loading', () =>
     {
         return SodiumHelper.init();
     });
@@ -107,7 +107,7 @@ describe ('Test of Stoa API Server', () =>
                 assert.strictEqual(response.data[0].address,
                     "GDNODE4KTE7VQUHVBLXIGD7VEFY57X4XV547P72D37SDG7UEO7MWOSNY");
                 assert.strictEqual(response.data[0].preimage.distance, null);
-            })
+            });
     });
 
     it ('Test of the path /validator - Wrong Height', () =>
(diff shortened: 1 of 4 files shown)


MichaelKim20:
I couldn't proceed without modifying the existing test code when the structure of the block was changed in the process of solving the issue (https://github.com/bpfkorea/stoa/issues/222).

tests/Stoa.test.ts
```diff
@@ -64,7 +64,7 @@ describe ('Test of Stoa API Server', () =>
         return stoa_server.stop().then(() => { return agora_server.stop() });
     });
 
-    it ('Test of the path /block_externalized', (doneIt: () => void) =>
+    before ('Test of the path /block_externalized', (doneIt: () => void) =>
```
Geod24: This looks wrong, it's not using promises and `before` doesn't make sense here.

tests/Stoa.test.ts
```diff
@@ -324,48 +298,42 @@ describe ('Test of Stoa API Server', () =>
...
-    it ('Test of the path /transaction_received', (doneIt: () => void) =>
+    describe ('Test of pending transactions', () =>
     {
-        let uri = URI(host)
-            .port(port)
-            .directory("transaction_received");
-
-        let url = uri.toString();
-        assert.doesNotThrow(async () =>
+        it ('Store one pending transaction', (doneIt: () => void) =>
```
Geod24, line 303: This is not asynchronous either ? It looks like a `before` though.

tests/Stoa.test.ts
```diff
@@ -324,48 +298,42 @@ describe ('Test of Stoa API Server', () =>
...
-            .catch((error) =>
-            {
-                assert.ok(!error, error);
-            })
-            .finally(doneIt);
+        it ('Test of the path /transactions/pending/:address', () =>
+        {
+            let uri = URI(host)
+                .port(port)
+                .directory("transactions/pending")
```
Geod24, line 322: How did this work before?

tests/Stoa.test.ts
MichaelKim20: All the tests that follow this are subordinate to this.
I changed it because I thought it would be a good idea to run the tests individually.
For now, I change it to the original one.

tests/Stoa.test.ts
MichaelKim20: Of course, it runs asynchronously, strictly speaking.
I maintained the existing method for the following reasons.
After this test is complete, the following tests must be carried out: I realized that I had to wait for the server to fully store the data, not completely after sending all the data. So I had no choice but to use `setTimeout` and `doneIt`.

tests/Stoa.test.ts
MichaelKim20: In the previous code, we found that the internal code was not normally checked.

The link below provides a detailed description and provides a solution to the same situation as ours.
https://www.testim.io/blog/testing-promises-using-mocha/ 
[Section: Why Mocha Struggles With Async Tasks]

MichaelKim20:
@Geod24 Thanks for the review, I have completed all your questions and requests. Please review it.

MichaelKim20:
The test code needs to be stabilized more.
So we need to think more. So I'll close this PR.

Geod24:
Maybe the last 3 commits could be made in their own PR ?

PR #256, opened 2021-01-18
Add the ability to search by a hash of the address [/utxo/:address]

MichaelKim20:
I added the ability to search for the hash value of the address on the endpoint that provides UTXO when the type of lock is KeyHash.


MichaelKim20:
This PR doesn't make sense, so I'll close it.

PR #262, opened 2021-01-27
Endpoint: Implement endpoint that provide the transaction status

MichaelKim20:
I implemented endpoints that provide the processing status of transactions.
It was added to check the transactions sent during the voting process and the wallet.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -1318,4 +1318,66 @@ export class LedgerStorage extends Storages
 
         return this.query(sql, [address]);
     }
+
+    /**
+     * Provides a status of a transaction.
+     * @param tx_hash The hash of the transaction
+     */
+    public getTransactionStatus (tx_hash: string): Promise<any>
+    {
+        let hash = new Hash(tx_hash).toBinary(Endian.Little);
+
+        let sql_tx =
+            `SELECT
+                B.hash,
+                T.block_height as height,
+                T.tx_hash
+            FROM
+                blocks B
+                INNER JOIN transactions T ON (B.height = T.block_height and T.tx_hash = ?);`;
+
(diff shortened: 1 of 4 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -1318,4 +1318,66 @@ export class LedgerStorage extends Storages
...
+                        resolve(result);
+                    }
+                    else
+                    {
+                        result.status = "not found";
+                        result.tx_hash = hash;
+                        resolve(result);
+                    }
+                }
+            }
```
TrustHenry, line 1376: How about moving this part to Stoa?

src/modules/storage/LedgerStorage.ts
```diff
@@ -1318,4 +1318,66 @@ export class LedgerStorage extends Storages
...
+                blocks B
+                INNER JOIN transactions T ON (B.height = T.block_height and T.tx_hash = ?);`;
+
+        let sql_tx_pending =
+            `SELECT
+                T.tx_hash
+            FROM
+                transaction_pool T
+            WHERE
+                T.tx_hash = ?;`;
```
TrustHenry, line 1345: I think we can combine the queries into one.

src/modules/storage/LedgerStorage.ts
```diff
@@ -1318,4 +1318,66 @@ export class LedgerStorage extends Storages
...
+
+        let result: any = {};
+        return new Promise<any>(async (resolve, reject) =>
+        {
+            try
+            {
+                let rows = await this.query(sql_tx_pending, [hash])
+                if (rows.length > 0)
+                {
+                    result.status = "pending";
```
TrustHenry, line 1355: We can define this code.

TrustHenry (approved):

PR #270, opened 2021-01-29
Revert "Lock/Unlock: Adapt Stoa to changes to Transaction, Input, Output

MichaelKim20:
Revert Lock/Unlock: Adapt Stoa to changes to Transaction, Input, Output


MichaelKim20:
Merges that have already been reviewed before.

Issue #274, opened 2021-02-08
Add ability to provide appropriate transaction fee


MichaelKim20:
Transaction fees are calculated by the following formula:
Fee(tx) = 0.00002 * Size(tx), If Fee(tx) is less than 0.01, choose 0.01.
Size(tx) is the number of bytes in the transaction.
<img width="797" alt="" src="https://user-images.githubusercontent.com/11639939/107174901-c4df7f80-6a0e-11eb-8c6a-32dc697a94cf.png">

- [x] SDKs should have the ability to measure the size of transactions.
- [x] Statistical work should be implemented to Stoa to provide transaction fees for Low, Middle, and High speeds.
- [x] An endpoint should be added to Stoa that provides transaction fees for Low, Middle, and High speeds.

PR #275, opened 2021-02-10
Add a shell endpoint that provides fees for the transaction

MichaelKim20:
I added a shell endpoint that provides fees for the transaction
The endpoint will be updated after implementing the ability to find appropriate fees.

Relates to #274


src/Stoa.ts
```diff
@@ -114,6 +115,7 @@ class Stoa extends WebService
         this.app.get("/transaction/pending/:hash", this.getTransactionPending.bind(this));
         this.app.get("/utxo/:address", this.getUTXO.bind(this));
         this.app.get("/transaction/status/:hash", this.getTransactionStatus.bind(this));
+        this.app.get("/transaction/fees/:size", this.getTransactionFees.bind(this));
```
```suggestion
        this.app.get("/transaction/fees/:tx_size", this.getTransactionFees.bind(this));
```
It's easier to understand.

src/Stoa.ts
```diff
@@ -366,6 +368,51 @@ class Stoa extends WebService
...
+     * Called when a request is received through the `/transaction/fees/:size` handler
+     *
+     * Returns transaction fees by the transaction size.
+     */
+    private getTransactionFees (req: express.Request, res: express.Response)
+    {
+        // TODO This is still incomplete.
+        //  We will proceed with statistics to find appropriate fees and then upgrade.
+
+        let size: string = req.params.size.toString();
```
TrustHenry: How about inputting the size as a `number`?

src/Types.ts
```diff
@@ -318,3 +318,29 @@ export class ConvertTypes
...
+
+/**
+ * Define the interface of the fee of the transaction
+ */
+export interface ITransactionFee
+{
+    /**
+     * The size of the transaction
+     */
+    size: number;
```
TrustHenry:
```suggestion
    tx_size: number;
```

TrustHenry:
We can get the average transaction fee from the past blocks, not the shell function.

MichaelKim20:
> We can get the average transaction fee from the past blocks, not the shell function.

Yes, that's right, I'll proceed with that next time.

src/Stoa.ts
```diff
@@ -366,6 +368,51 @@ class Stoa extends WebService
...
+        // TODO This is still incomplete.
+        //  We will proceed with statistics to find appropriate fees and then upgrade.
+
+        let size: string = req.params.tx_size.toString();
+
+        logger.http(`GET /transaction/fees/${size}}`);
+
+        if (!Utils.isPositiveInteger(size))
+        {
+            res.status(400).send(`Invalid value for parameter 'size': ${size}`);
```
```suggestion
            res.status(400).send(`Invalid value for parameter 'tx_size': ${size}`);
```

src/Stoa.ts
MichaelKim20: In the following row, check to see if it is a positive number. 
It is only possible when it is a string

TrustHenry (approved):
LGTM

PR #279, opened 2021-02-17
Wallet: Modify the filtering function of the Transaction Type

MichaelKim20:
If a special filter exists, payload and freeze are excluded, and all types of transactions are returned only when there is no filter.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -623,8 +623,7 @@ class Stoa extends WebService
 
         filter_type = (req.query.type !== undefined)
             ? req.query.type.toString().split(',').map((m) => ConvertTypes.toDisplayTxType(m))
-            : [0, 1];
-        filter_type.push(...[DisplayTxType.Freeze, DisplayTxType.Payload].filter(n => filter_type.find(m => m == n) === undefined));
+            : [0, 1, 2, 3];
 
         if (filter_type.find(m => (m === -1)) !== undefined)
         {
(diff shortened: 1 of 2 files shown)


src/Stoa.ts
```diff
@@ -623,8 +623,7 @@ class Stoa extends WebService
 
         filter_type = (req.query.type !== undefined)
             ? req.query.type.toString().split(',').map((m) => ConvertTypes.toDisplayTxType(m))
-            : [0, 1];
-        filter_type.push(...[DisplayTxType.Freeze, DisplayTxType.Payload].filter(n => filter_type.find(m => m == n) === undefined));
+            : [0, 1, 2, 3];
```
MichaelKim20, line 626: It includes all conditions when there is no filter.

TrustHenry (approved):

src/Stoa.ts
```diff
@@ -623,8 +623,7 @@ class Stoa extends WebService
 
         filter_type = (req.query.type !== undefined)
             ? req.query.type.toString().split(',').map((m) => ConvertTypes.toDisplayTxType(m))
-            : [0, 1];
-        filter_type.push(...[DisplayTxType.Freeze, DisplayTxType.Payload].filter(n => filter_type.find(m => m == n) === undefined));
```
MichaelKim20, line 627: In all cases, we removed the addition of freeze and payload.

tests/WalletAPI.test.ts
```diff
@@ -310,4 +310,30 @@ describe ('Test of Stoa API for the wallet with `sample_data`', () => {
 
         assert.deepStrictEqual(expected, response.data);
     });
+
+    it ('Test of the path /wallet/transactions/history - Filtering - exclude DataPayload in specific filter', async () =>
```
MichaelKim20, line 314: There was one when it was `payload`, but nothing is returned when it is `outbound`.

MichaelKim20:
@TrustHenry Thank for your review

Issue #280, opened 2021-02-23
Apply Agora's change to Stoa to prevent the refund output from freezing.


MichaelKim20:
Agora was modified by the PR below.
https://github.com/bpfkorea/agora/pull/1664
It is necessary to apply these changes to Stoa and SDK.

PR #281, opened 2021-02-25
Prevent freezing of refund output in freeze transactions

MichaelKim20:
I implemented the processing in the same way as Agora to prevent the refund output from freezing in a freeze transaction.

Relates to https://github.com/bpfkorea/stoa/issues/280

### DIFF
--- a/docs/Database_Schema.md
+++ b/docs/Database_Schema.md
@@ -168,6 +168,8 @@ CREATE TABLE IF NOT EXISTS "tx_outputs" (
 |:----------------- |:--------- |:--:|:--------:| -------- | --------- |
 |  utxo_key         | BLOB      | Y  | Y        |          | The hash of the UTXO|
 |  tx_hash          | BLOB      |    | Y        |          | The hash of transaction |
+|  type             | INTEGER   |    | Y        |          | The type of UTXO (0: Payment, 1: Freeze) If the type of transaction is `Freeze` and the refund output is less than 40,000 BOA, it is `Payment`. Others are the same as the transaction type. |
+|  unlock_height    | INTEGER   |    | Y        |          | Height of the block to be unlock|
 |  amount           | NUMERIC   |    | Y        |          | The monetary value of this output, in 1/10^7|
 |  lock_type        | INTEGER   |    | Y        |          | (0: Key; 1: Hash of Key; 2: Script; 3: Hash of Script) |
 |  lock_bytes       | BLOB      |    | Y        |          | The bytes of lock |
@@ -179,6 +181,8 @@ CREATE TABLE IF NOT EXISTS "tx_outputs" (
 CREATE TABLE IF NOT EXISTS "utxos" (
     "utxo_key"              BLOB    NOT NULL,
     "tx_hash"               BLOB    NOT NULL,
+    "type"                  INTEGER NOT NULL,
+    "unlock_height"         INTEGER NOT NULL,
     "amount"                NUMERIC NOT NULL,
     "lock_type"             INTEGER NOT NULL,
     "lock_bytes"            BLOB    NOT NULL,
--- a/docs/Database_Schema.md
+++ b/docs/Database_Schema.md
@@ -141,7 +141,6 @@ CREATE TABLE IF NOT EXISTS "tx_inputs" (
 |  lock_type        | INTEGER   |    | Y        |          | (0: Key; 1: Hash of Key; 2: Script; 3: Hash of Script) |
 |  lock_bytes       | BLOB      |    | Y        |          | The bytes of lock |
 |  address          | TEXT      |    | Y        |          | The public key, Valid only when lock type is 0. Other than that, it's a blank.|
-|  used             | INTEGER   |    | Y        | 0        | Whether this output was used or not(1: used, 0: not used)|
 
 ### _Create Script_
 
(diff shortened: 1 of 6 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -589,8 +590,49 @@ export class LedgerStorage extends Storages
...
+                    let sql =
+                        `SELECT
+                            count(*) as count
+                        FROM
+                            blocks B
+                            INNER JOIN transactions T ON (B.height = T.block_height)
+                            INNER JOIN tx_outputs O ON (T.tx_hash = O.tx_hash)
+                        WHERE
+                            O.type = 1
+                            and O.utxo_key in (${utxo.join(',')})
```
```suggestion
                        `SELECT
                            count(*) as count
                        FROM
                            blocks B
                            INNER JOIN transactions T ON (B.height = T.block_height)
                            INNER JOIN tx_outputs O ON (T.tx_hash = O.tx_hash)
                        WHERE
                            T.type = 1
                            and O.used = 0
                            and O.utxo_key in (${utxo.join(',')})
```
Do you need it?

docs/Database_Schema.md
```diff
@@ -152,6 +154,8 @@ CREATE TABLE IF NOT EXISTS "tx_outputs" (
     "output_index"          INTEGER NOT NULL,
     "tx_hash"               BLOB    NOT NULL,
     "utxo_key"              BLOB    NOT NULL,
+    "type"                  INTEGER NOT NULL,
```
TrustHenry, line 184: If this is a transaction type, you can refer to the transaction table.

MichaelKim20: The conditions below were added when recording TxOutput.
```TypeScript
                let unlock_height: JSBI;
                let tx_type: TxType;
                if (melting && address != TxPayloadFee.CommonsBudgetAddress)
                {
                    tx_type = tx.type;
                    unlock_height = JSBI.add(height.value, JSBI.BigInt(2016));
                }
                else if (tx.type == TxType.Freeze && tx.outputs.length >= 2 &&
                    (out_idx == 0 && JSBI.lessThan(output.value, JSBI.BigInt(400_000_000_000))))
                {
                    tx_type = TxType.Payment;
                    unlock_height = JSBI.add(height.value, JSBI.BigInt(1));
                }
                else
                {
                    tx_type = tx.type;
                    unlock_height = JSBI.add(height.value, JSBI.BigInt(1));
                }
```

src/modules/storage/LedgerStorage.ts
MichaelKim20: I created the function `is_melting`

docs/Database_Schema.md
MichaelKim20: The type of table tx_outputs can have a different value than the type of table transactions.
The type of transaction in the transactions type.
This function was added from Agora to prevent the refund output of the transaction from being frozen.
Agora has a type in UTXO, and I added it to the table.

src/modules/storage/LedgerStorage.ts
MichaelKim20: This is necessary for the issue to be resolved.

The type of table tx_outputs can have a different value than the type of table transactions.
The type of transaction in the transactions type.
This function was added from Agora to prevent the refund output of the transaction from being frozen.
Agora has a type in UTXO, and I added it to the table.

docs/Database_Schema.md
TrustHenry: Oh, I think we will make a separate UTXOs table.
I think the update query will happen again and again.
Frequent updates can cause lock or deadlocks.
We recommend that you use the Transactions table as read-only.

src/modules/storage/LedgerStorage.ts
TrustHenry: `and O.used = 0` This condition is required.
However, it can be changed according to the comments being discussed below.

MichaelKim20: It has been modified by adding a table that only manages UTXO.

docs/Database_Schema.md
MichaelKim20: It has been modified by adding a table that only manages UTXO.

src/modules/storage/LedgerStorage.ts
```diff
@@ -569,13 +580,12 @@ export class LedgerStorage extends Storages
...
 
-        function update_spend_output (storage: LedgerStorage,
-            input: TxInput): Promise<void>
+        function update_spend_output (storage: LedgerStorage, input: TxInput): Promise<void>
         {
             return new Promise<void>((resolve, reject) =>
             {
                 storage.run(
-                    `UPDATE tx_outputs SET used = 1 WHERE utxo_key = ?`,
+                    `DELETE FROM utxos WHERE utxo_key = ?`,
```
TrustHenry, line 588: Nice! 
What should We do if there is an error in this part?

src/modules/storage/LedgerStorage.ts
```diff
@@ -569,13 +580,12 @@ export class LedgerStorage extends Storages
             });
         }
 
-        function update_spend_output (storage: LedgerStorage,
-            input: TxInput): Promise<void>
+        function update_spend_output (storage: LedgerStorage, input: TxInput): Promise<void>
```
```suggestion
        function delete_spend_output (storage: LedgerStorage, input: TxInput): Promise<void>
```

docs/Database_Schema.md
```diff
@@ -156,13 +155,44 @@ CREATE TABLE IF NOT EXISTS "tx_outputs" (
...
-## 6. Table **validators**
+## 6. Table **utxos**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+|  utxo_key         | BLOB      | Y  | Y        |          | The hash of the UTXO|
+|  tx_hash          | BLOB      |    | Y        |          | The hash of transaction |
+|  type             | INTEGER   |    | Y        |          | The type of transaction |
```
TrustHenry: We think we need a more detailed explanation to get rid of the confusion later.

TrustHenry (approved):
LGTM, Except for the comments.

src/modules/storage/LedgerStorage.ts
TrustHenry: For example, there is no utxo to erase, or there is throw error.

MichaelKim20: That's a good point. I'm gonna have to test it out.

MichaelKim20: When running a query using a non-existent UTXO, no error occurred as follows:
```
Execution finished without errors.
Result: query executed successfully. Took 0ms, 0 
At line 1:
DELETE FROM utxos WHERE utxo_key = x'4FAB6478E5283258DD749EDCB303E48F4192F199D742E14B348711F4BBB116B197E63429C6FA608621681E625BAF1B045A07ECF12F2E0B04C38BEE449F5EACFA'
```

Issue #285, opened 2021-03-08
The function to check transactions on the Votera server is needed.


MichaelKim20:
The ability to provide transactions to specific addresses over a specific period of time is required. This is necessary to confirm the deposit of the proposal's fees and voting costs.

PR #286, opened 2021-03-08
Endpoint: Add an endpoint /transaction:hash

MichaelKim20:
This is an endpoint that provides information about a confirmed transaction.

Relates to #285


MichaelKim20:
This is similar to the "endpoint /transaction/pending/:hash" that is already implemented, so I merge it after approval.

PR #287, opened 2021-03-08
Bug Fix: Change to use Lock instead of address

MichaelKim20:
When looking up a pending transaction, it should be changed to Lock instead of address.


TrustHenry (approved):

MichaelKim20:
Thank you

PR #290, opened 2021-03-15
Wallet: Add delayed block height to the endpoint that provide pending transactions

### LINKED ISSUE #289: Values for `Block delay` blocks for pending transactions for Wallet shall be provided.
Currently, Wallet calculates using the transaction's receive time.
Rather than this method, we use a method that utilizes block heights and current block heights when receiving transactions to compute and deliver them from Stoa.
This prevents problems from occurring even if the value of the time stamp in the block changes to a time offset in Stoa.

MichaelKim20:
I made this to solve issue #289 

Relates to #289


docs/Database_Schema.md
```diff
@@ -296,6 +297,7 @@ CREATE TABLE IF NOT EXISTS "transaction_pool" (
     "type"                  INTEGER NOT NULL,
     "payload"               BLOB    NOT NULL,
     "lock_height"           INTEGER NOT NULL,
+    "receive_height"        INTEGER NOT NULL,
```
TrustHenry:
```suggestion
    "received_height"        INTEGER NOT NULL,
```

TrustHenry (approved):
LGTM, I wish the word could be changed.

PR #291, opened 2021-03-16
Adapted Stoa to change from `BlockHeader.timestamp` to `time_offset'

### LINKED ISSUE #284: Adapt Stoa to change from `BlockHeader.timestamp` to `BlockHeader.time_offset`
https://github.com/bpfkorea/agora/pull/1693

TrustHenry:
Calculates the block timestamp by adding offset based on the 
Genesis timestamp of the consensus config.

Related to #284


src/modules/storage/LedgerStorage.ts
```diff
@@ -17,6 +17,7 @@ import {
     Utils, Endian, Lock, Unlock, PublicKey, DataPayload, TxPayloadFee
 } from 'boa-sdk-ts';
 import { Storages } from './Storages';
+import { Config } from '../../modules/common/Config';
```
```suggestion
import { Config } from '../common/Config';
```

src/Stoa.ts
```diff
@@ -44,17 +45,23 @@ class Stoa extends WebService
...
+
     /**
      * Constructor
      * @param database_filename sqlite3 database file name
      * @param agora_endpoint The network endpoint to connect to Agora
      * @param port The network port of Stoa
      * @param address The network address of Stoa
      */
-    constructor (database_filename: string, agora_endpoint: URL, port: number | string, address: string)
+    constructor (config: Config, database_filename: string, agora_endpoint: URL, port: number | string, address: string)
```
MichaelKim20: We did not use Config directly inside Stoa to avoid creating a dependency between Class Config and Class Stoa.
Therefore, it would be a little uncomfortable, but it would be better to deliver only the necessary variables.
```suggestion
    constructor (database_filename: string, agora_endpoint: URL, port: number | string, address: string, genesis_timestamp: number)
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -25,21 +26,27 @@ import JSBI from 'jsbi';
...
+    /**
+     * Main config
+     */
+    private config: Config;
+
     /**
      * Construct an instance of `LedgerStorage`, exposes callback API
      */
-    constructor (filename: string, callback: (err: Error | null) => void)
+    constructor (filename: string, config: Config, callback: (err: Error | null) => void)
```
MichaelKim20: It would be better to deliver only the necessary variables.

MichaelKim20 (changes requested):
The method you used is also good. My opinion is that delivering only the variables used will reduce errors during testing and operation.
The default value is being used for Config created by the test code.
If the default value is changed, it will be difficult to find the cause of the error easily in the test code.
The rest is very good.

TrustHenry:
@MichaelKim20 Thanks for the reviews, I apply the requested changes.

src/modules/storage/LedgerStorage.ts
```diff
@@ -17,6 +17,7 @@ import {
     Utils, Endian, Lock, Unlock, PublicKey, DataPayload, TxPayloadFee
 } from 'boa-sdk-ts';
 import { Storages } from './Storages';
+import { Config } from '../common/Config';
```
```suggestion
```
I don't think we need this.

src/modules/storage/LedgerStorage.ts
```diff
@@ -765,13 +776,13 @@ export class LedgerStorage extends Storages
...
-                            amount, 
-                            lock_type, 
-                            lock_bytes, 
+                            utxo_key,
+                            tx_hash,
+                            type,
+                            unlock_height,
+                            amount,
+                            lock_type,
+                            lock_bytes,
```
MichaelKim20, line 784: Good catch.

MichaelKim20 (approved):
It's very good.
Please correct one last thing.

TrustHenry:
@MichaelKim20 Thank you.

Issue #296, opened 2021-03-23
SPV: Add an endpoint that takes `tx_hash` and returns `block_height` and `merkle_root`


kchulj:
The purpose of this method is to check whether the provided `tx_hash` exists in the blockchain.
If it exists, then return the corresponding `block_height`, and the `merkle_root` of the block of that height.

Part of [agora#239](https://github.com/bosagora/agora/issues/239)

PR #297, opened 2021-03-24
Implement `getBlockHeaderByTxHash` function

kchulj:
The purpose of this method is to check whether the provided `tx_hash` exists in the blockchain.
If it exists, then return the corresponding `block_height`, and the `merkle_root` of the block of that height.

Solution for #296
Part of [agora#239](https://github.com/bosagora/agora/issues/239)

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -147,7 +147,7 @@ export class LedgerStorage extends Storages
             address             TEXT    NOT NULL,
             PRIMARY KEY(utxo_key)
         );
-        
+
         CREATE TABLE IF NOT EXISTS payloads (
             tx_hash             BLOB    NOT NULL,
             payload             BLOB    NOT NULL,
@@ -233,7 +233,7 @@ export class LedgerStorage extends Storages
     public putBlocks (block: Block): Promise<void>
     {
         let genesis_timestamp: number = this.genesis_timestamp;
-        
+
         function saveBlock (storage: LedgerStorage, block: Block, genesis_timestamp: number): Promise<void>
         {
             return new Promise<void>((resolve, reject) =>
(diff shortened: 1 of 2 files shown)


MichaelKim20:
I'd love to.
Don't you need any more information?

MichaelKim20 (approved):
LGTM

src/modules/storage/LedgerStorage.ts
```diff
@@ -1806,4 +1806,28 @@ export class LedgerStorage extends Storages
...
+        `SELECT
+            height, merkle_root
+        FROM
+            transactions T
+        INNER JOIN
+            blocks B
+        ON
+            T.block_height = B.height
+        AND
+            T.tx_hash = ?`;
```
TrustHenry, line 1829:
```suggestion
        `SELECT
            height, merkle_root
        FROM
            transactions T
            INNER JOIN blocks B ON T.block_height = B.height
        AND
            T.tx_hash = ?`;
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -1806,4 +1806,28 @@ export class LedgerStorage extends Storages
...
+     * @param tx_hash The hash of the transaction
+     * @returns Returns the Promise. If it is finished successfully the `.then`
+     * of the returned Promise is called with the records
+     * and if an error occurs the `.catch` is called with an error.
+     */
+    public getBlockHeaderByTxHash (tx_hash: Hash): Promise<any[]>
+    {
+        let sql =
+        `SELECT
+            height, merkle_root
```
TrustHenry, line 1821: How about including all the block header information?

kchulj: Yes, since the name is `getBlockHeaderByTxHash`.
@MichaelKim20 what do you think?

MichaelKim20: Okay 
Let's add what we need next time.

TrustHenry (approved):
LGTM

src/modules/storage/LedgerStorage.ts
kchulj: Oh I didn't apply this style yet :)

TrustHenry: We'll add more next time. :)

kchulj: Thank you~

PR #303, opened 2021-04-01
Switch address format from Stellar-style to bech32

MichaelKim20:
Recently, the way Agora's public address is printed as a string has been changed.
I adapted to Stoa for this change.

Related to bosagora/agora#1655


TrustHenry (approved):
LGTM

Issue #300, opened 2021-03-26
SPV: Implement method `checkMerklePath`


kchulj:
We need the Merkle proof to check whether the transaction is contained in the block.

The function in agora is `public static Hash checkMerklePath` which returns the Merkle root, but it could be `bool` here.
Open to suggestions.

Part of [agora#239](https://github.com/bosagora/agora/issues/239)

PR #306, opened 2021-04-04
SPV: Implement endpoint `verifyPayment`

kchulj:
- `checkMerklePath` takes `merkle_path`, `tx_hash`, `tx_index` and returns the Merkle root, as in Agora.

- `verifyPayment` is the main function for the SPV. It checks whether the provided transaction is included in a block.

Solution for #300 / [agora#239](https://github.com/bosagora/agora/issues/239)


MichaelKim20 (approved):
Nice!

PR #308, opened 2021-04-05
Trivial: Fix doc

kchulj:


src/modules/storage/LedgerStorage.ts
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    The class that create and insert and read the ledger into the database.
+    The class that creates, inserts, and reads the ledger into the database.
```
hewison-chris: You have an extra comma.

hewison-chris (changes requested):
Small change otherwise good.

src/modules/storage/LedgerStorage.ts
kchulj: Fixed :)

hewison-chris (approved):
LGTM

TrustHenry (approved):

PR #310, opened 2021-04-06
Agora sometimes responds late to requests

TrustHenry:


src/modules/agora/AgoraClient.ts
```diff
@@ -65,7 +65,7 @@ export class AgoraClient implements FullNodeAPI
         this.client = axios.create({
             baseURL: endpoint.toString(),
             // Timeout are in ms, so 2s timeout
```
kchulj: Should we change comment too?

kchulj: Since we may change the value again, I think we can remove.
```suggestion
            // Timeout is in ms
```

kchulj (changes requested):

src/modules/agora/AgoraClient.ts
TrustHenry: Good eye! Thanks.

PR #311, opened 2021-04-06
Hash: Change the method of calculating hash values to the same as Agora

MichaelKim20:
Recently, the method of calculating the hash in Agora has been changed.
As a result, transactions are not being processed normally when they are sent from the wallet.
The newly added code included the size type of the array and the length of the array when the array was hash.
I changed the hash value calculation method to the same as Agora.

Related to https://github.com/bosagora/boa-sdk-ts/pull/146


MichaelKim20:
I changed the data and modified the hash values accordingly.
I am planning to merge first and conduct the test with Agora.
If there is a problem later, I will correct it.

PR #313, opened 2021-04-07
Add a workflow to build the Docker image

Geod24:


Geod24:
We won't see the result until we merge, so self-merging and checking if it fixes the issue (we haven't pushed a docker image in 20 days).

Geod24:
It worked \o/

PR #314, opened 2021-04-08
Fix: Replace with 0 when the result of SUM in SQL is NULL

MichaelKim20:
I fix the bug.
I replaced with 0 when the result of SUM in SQL is NULL.


TrustHenry (approved):
LGTM

kchulj (approved):

PR #324, opened 2021-05-06
Wallet: Remove unnecessary errors(HTTP Status 204).

### LINKED ISSUE #323: When there is no data in any transaction history it gives axios error
Axios generates an error if the HTTP Status code is not number 200.

MichaelKim20:
I think it is reasonable to return the Http Status Code to 200 because the number of this service can be zero even if it is normal.


TrustHenry:
It would be better if the wallet (client) has 204 status code processing classified.
But if this is a problem, I can erase it now.

TrustHenry (approved):

PR #329, opened 2021-05-10
Fix HTTP response code passed to invalid values

MichaelKim20:
Change to 400 which means Bad Request.


kchulj (approved):

MichaelKim20:
@kchulj Thank you

PR #346, opened 2021-06-07
Set tx type for each tx output

MichaelKim20:
It was applied in accordance with the change in Agora.
The tx_type in Transaction was moved to each output.
This will ensure that refunding outputs are not frozen in freeze transactions.

https://github.com/bosagora/boa-sdk-ts/pull/178
Related to https://github.com/bosagora/agora/pull/2114


TrustHenry (approved):
LGTM

MichaelKim20:
Thank you for review

PR #352, opened 2021-06-08
Replace `DataPayload` with `Buffer` (BOA-SDK-TS Ver 0.0.48)

MichaelKim20:
It was applied in accordance with the change in Agora. https://github.com/bosagora/agora/pull/2138
It changed the transaction member, payload, to an instance of Buffer.


TrustHenry (approved):
LGTM

PR #362, opened 2021-06-14
Add `TransactionPool` for double-spend transaction processing

MichaelKim20:
This is necessary when a transaction is canceled, to reflect it in a pending transaction.

### DIFF
--- a/docs/Database_Schema.md
+++ b/docs/Database_Schema.md
@@ -448,3 +448,22 @@ Create TABLE IF NOT EXISTS "marketcap" (
     PRIMARY KEY ("last_updated_at")
 );
 ```
+
+## 16. Table **tx_pool**
+
+### _Schema_
+
+| Column            | Data Type | PK | Not NULL | Default  |Description|
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+| key               | TINYBLOB  | Y  | Y        |          | The hash of transaction |
+| val               | BLOB      |    | Y        |          | The transaction serialized to binary |
+
+### _Create Script_
+
+```sql
+CREATE TABLE IF NOT EXISTS tx_pool (
+    `key`   TINYBLOB    NOT NULL,
+    `val`   BLOB        NOT NULL,
+    PRIMARY KEY (`key`(64))
+)
+```
--- a/src/modules/storage/TransactionPool.ts
+++ b/src/modules/storage/TransactionPool.ts
@@ -1,0 +1,224 @@
+/*******************************************************************************
+
+    Contains a transaction pool that is serializable to disk,
+    using MySQL as a store.
+    It was added to remove double-spending transactions that use the same input.
+
+    Copyright:
+        Copyright (c) 2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+import { Endian, Hash, hashFull, Transaction } from "boa-sdk-ts";
+import { SmartBuffer } from "smart-buffer";
+
+import * as mysql from "mysql2";
+
+/**
+ * It was added to remove double-spending transactions that use the same input.
+ */
+export class TransactionPool
+{
+    /**
+     * Keeps track of which TXs spend which inputs
+     */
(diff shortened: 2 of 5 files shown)


docs/Database_Schema.md
```diff
@@ -448,3 +448,22 @@ Create TABLE IF NOT EXISTS "marketcap" (
...
+|:----------------- |:--------- |:--:|:--------:| -------- | --------- |
+| hash              | TEXT      | Y  | Y        |          | The hash of transaction|
+| tx                | BLOB      | Y  | Y        |          | The transaction serialized to binary |
+
+### _Create Script_
+
+```sql
+CREATE TABLE IF NOT EXISTS tx_pool (
+    hash   TEXT    NOT NULL,
+    tx     BLOB    NOT NULL,
```
TrustHenry:
```suggestion
    tx_hash_text   TEXT    NOT NULL,
    tx_bytes     BLOB    NOT NULL,
```

src/Stoa.ts
```diff
@@ -1462,6 +1480,9 @@ class Stoa extends WebService
                         {
                             if (JSBI.equal(block.header.height.value, expected_height.value))
                             {
+                                for (let tx of block.txs)
+                                    await this.transaction_pool.remove(tx);
```
TrustHenry: You can move the call to the point. `ledger_storage.putBlocks `

src/Stoa.ts
```diff
@@ -23,6 +24,7 @@ import { URL } from 'url';
 class Stoa extends WebService
 {
     private _ledger_storage: LedgerStorage | null;
+    private _transaction_pool: TransactionPool | null;
```
TrustHenry: You can move to `ledger_storage`

src/Stoa.ts
MichaelKim20: Yeah, I was thinking about that, too. Now that we've found a good way, we'd better do it.

docs/Database_Schema.md
MichaelKim20: It's not too long. I don't think it's necessary to describe the form.

MichaelKim20: I'd like to prohibit the use of this table by joining other tables.

docs/Database_Schema.md
```diff
@@ -448,3 +448,22 @@ Create TABLE IF NOT EXISTS "marketcap" (
...
 );
 ```
+
+## 16. Table **tx_pool**
+
+### _Schema_
+
+| hash              | TEXT      | Y  | Y        |          | The hash of transaction|
```
TrustHenry:
```suggestion
| hash              | TEXT      | Y  | Y        |          | The text-encoded hash of transaction |
```

docs/Database_Schema.md
TrustHenry: Then it can replace it with an explanation.

docs/Database_Schema.md
MichaelKim20: I changed the name to be the same as Agora, please review it again.

src/modules/storage/TransactionPool.ts
```diff
@@ -0,0 +1,222 @@
...
+        for (const input of tx.inputs)
+        {
+            const in_hash_string = hashFull(input).toString();
+            // Update the spenders list
+            if (!this.spenders.has(in_hash_string))
+                this.spenders.set(in_hash_string, new Set<string>());
+
+            let set = this.spenders.get(in_hash_string);
+            if ((set !== undefined) && (!set.has(tx_hash_string)))
+                set.add(tx_hash_string);
```
TrustHenry: ?

TrustHenry: Please check it.

MichaelKim20: > Please check it.

```
            let set = this.spenders.get(in_hash_string);
            if  (!set.has(tx_hash_string))
                set.add(tx_hash_string);
```
This results in an error because the `set` can be `undefined`

MichaelKim20: @TrustHenry I changed it to a simpler code. Check it out, please.

TrustHenry (approved):

PR #363, opened 2021-06-14
Add endpoints that provide information about UTXO for multiple UTXO hashs

MichaelKim20:
This is necessary to implement functions to cancel transactions that have already been sent.
In order to cancel a pending transaction, input information is needed.
This PR will provide that.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -134,6 +134,7 @@ class Stoa extends WebService
         this.app.get("/transaction/pending/:hash", this.getTransactionPending.bind(this));
         this.app.get("/transaction/:hash", this.getTransaction.bind(this));
         this.app.get("/utxo/:address", this.getUTXO.bind(this));
+        this.app.post("/utxos", this.getUTXOs.bind(this));
         this.app.get("/transaction/status/:hash", this.getTransactionStatus.bind(this));
         this.app.get("/transaction/fees/:tx_size", this.getTransactionFees.bind(this));
         this.app.get("/wallet/transactions/history/:address", this.getWalletTransactionsHistory.bind(this));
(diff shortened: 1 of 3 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -1509,6 +1509,43 @@ export class LedgerStorage extends Storages
...
+                O.amount,
+                O.lock_type,
+                O.lock_bytes,
+                T.block_height,
+                B.time_stamp as block_time,
+                O.type
+            FROM
+                tx_outputs O
+                INNER JOIN transactions T ON (T.tx_hash = O.tx_hash)
+                INNER JOIN blocks B ON (B.height = T.block_height)
```
TrustHenry: Why import from `outputs` without referencing the `utxos` table?

MichaelKim20: Because there may be cases where the data we are looking for is not in Table utxos.
On second thought, however, I think we need to let them know when we don't have it.
Therefore, I am planning to change the form of the return value to table utxos.

MichaelKim20:
@TrustHenry I changed code

src/Stoa.ts
```diff
@@ -586,6 +587,72 @@ class Stoa extends WebService
...
     }
 
+    /**
+     * POST /utxos
+     *
+     * Returns UTXO's information about the UTXO hash array.
+     *
+     * Returns a array of UTXO's information.
+     */
+    private getUTXOs (req: express.Request, res: express.Response)
```
TrustHenry: It would be great if the request parameter description is added.
`The array of UTXO hash`

src/Types.ts
```diff
@@ -96,6 +96,11 @@ export interface IUnspentTxOutput
      * The bytes of lock script in transaction output
      */
     lock_bytes: string;
+
+    /**
+     * Does the value exist in the UTXO table?
+     */
+    exist?: boolean
```
TrustHenry: The question is strange.
`Whether to include UTXO table`

TrustHenry (approved):
LGTM
It looks good except for two document descriptions.

src/Stoa.ts
MichaelKim20: I fixed

src/Types.ts
MichaelKim20: It was removed and treated as returning an array of utxo.

PR #369, opened 2021-06-22
Implement the ability to provide reasonable transaction fees

MichaelKim20:
Implement the ability to provide reasonable transaction fees
1. Each time a block is stored, it is recently calculated for 100 blocks.
2. Calculate the average disparity between fees calculated by the formula and actual fees.
3. Remove the top 5% and the bottom 5% to reduce error data from the average value calculation.
4. The calculation fee plus the average value of the disparity provides a three-step transaction fee.

### DIFF
--- a/src/modules/common/FeeManager.ts
+++ b/src/modules/common/FeeManager.ts
@@ -1,0 +1,76 @@
+/*******************************************************************************
+
+    This file contains a class that calculate transaction fees and
+    process statistics.
+
+    Copyright:
+        Copyright (c) 2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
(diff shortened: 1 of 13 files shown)


src/modules/common/FeeManager.ts
```diff
@@ -0,0 +1,76 @@
...
+            low = JSBI.BigInt(minimum);
+
+        return [high, medium, low];
+    }
+
+    /**
+     * Calculates an average of the disparity
+     * @param values The sample of disparity
+     */
+    public static calculateAvgDisparity (values: Array<number>): number
```
TrustHenry: Document not found for return value.

TrustHenry (approved):
Nice! LGTM

src/modules/common/FeeManager.ts
MichaelKim20: I added a documentation

PR #376, opened 2021-07-01
Apply cross-env POSIX, Windows both support

TrustHenry:
Environment variables in npm scripts are cross-OS supported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
     "description": "API server for Agora",
     "main": "src/main.js",
     "scripts": {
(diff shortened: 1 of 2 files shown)


kchulj (approved):

package.json
```diff
@@ -34,6 +34,7 @@
         "coingecko-api-v3": "0.0.11",
         "cors": "^2.8.5",
         "crc": "^3.8.0",
+        "cross-env": "^7.0.3",
```
MichaelKim20: Isn't it good to add to devDependencies?
npm install --save-dev cross-env

TrustHenry: npm start is also affected.

MichaelKim20 (approved):

PR #389, opened 2021-07-14
Update endpoint `/wallet/transactions/pending/:address`

MichaelKim20:
The amount of deposit and withdrawal are calculated and revised to aggregate the amount of the recipient.
The address is modified to the recipient's address

I also corrected the error below

image

Below is the data dump file of MYSQL that I used in the test.
[Stoa-Dump20210714.sql.zip](https://github.com/bosagora/stoa/files/6813288/Stoa-Dump20210714.sql.zip)


TrustHenry (approved):
LGTM

PR #394, opened 2021-07-21
Endpoint: Add an endpoint that provides the balance of the entered address

MichaelKim20:
I made a UTXO that only provides the balance without checking the entire UTXO.
Currently, all UTXO is required to check the balance of a specific address.
However, if this is added, the load on the balance calculation can be reduced.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -2382,6 +2382,52 @@ export class LedgerStorage extends Storages {
     }
 
     /**
+     * Provides a balance of address
+     * @param address The address to check the balance
+     */
+    public getWalletBalance(address: string): Promise<any[]> {
+        let sql = `
+            SELECT
+                ? as address,
+                IFNULL(SUM(amount), 0) AS balance,
+                IFNULL(SUM(CASE WHEN ((unlock_height <= height + 1) AND ((type = 0) OR (type = 2))) THEN amount ELSE 0 END), 0) AS spendable,
+                IFNULL(SUM(CASE WHEN ((unlock_height <= height + 1) AND ((type = 1))) THEN amount ELSE 0 END), 0) AS frozen,
+                IFNULL(SUM(CASE WHEN ((unlock_height > height + 1) AND ((type = 0) OR (type = 2))) THEN amount ELSE 0 END), 0) AS locked
+            FROM
+            (
+                SELECT
+                    O.utxo_key as utxo,
+                    O.amount,
+                    O.lock_type,
+                    O.lock_bytes,
+                    T.block_height,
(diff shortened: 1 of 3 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -2381,6 +2381,52 @@ export class LedgerStorage extends Storages {
...
     }
 
+    /**
+     * Provides a balance of address
+     * @param address The address to check the balance
+     */
+    public getWalletBalance(address: string): Promise<any[]> {
+        let sql = `
+            SELECT
+                ? as address,
```
TrustHenry, line 2391:
```suggestion
                T.address as address,
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -2381,6 +2381,52 @@ export class LedgerStorage extends Storages {
...
+                IFNULL(SUM(amount), 0) AS balance,
+                IFNULL(SUM(CASE WHEN ((unlock_height <= height + 1) AND ((type = 0) OR (type = 2))) THEN amount ELSE 0 END), 0) AS spendable,
+                IFNULL(SUM(CASE WHEN ((unlock_height <= height + 1) AND ((type = 1))) THEN amount ELSE 0 END), 0) AS frozen,
+                IFNULL(SUM(CASE WHEN ((unlock_height > height + 1) AND ((type = 0) OR (type = 2))) THEN amount ELSE 0 END), 0) AS locked
+            FROM
+            (
+                SELECT
+                    O.utxo_key as utxo,
```
TrustHenry, line 2399:
```suggestion
                    O.utxo_key as utxo,
                    O.address,
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -2381,6 +2381,52 @@ export class LedgerStorage extends Storages {
...
+                        FROM
+                            utxos S
+                            INNER JOIN tx_input_pool I ON (I.utxo = S.utxo_key)
+                            INNER JOIN transaction_pool T ON (T.tx_hash = I.tx_hash)
+                        WHERE
+                            S.address = ?
+                    )
+            ) AS T;`;
+
+        return this.query(sql, [address, address, address]);
```
TrustHenry, line 2427:
```suggestion
        return this.query(sql, [address, address]);
```

TrustHenry (approved):
LGTM

src/modules/storage/LedgerStorage.ts
MichaelKim20: When the data did not exist, the address came out null and was treated like that's

MichaelKim20:
@TrustHenry Thank you for review

PR #395, opened 2021-07-21
Fix the validators get query

TrustHenry:
Added height to group criteria in validators query.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -1912,33 +1912,29 @@ export class LedgerStorage extends Storages {
                 enrollments.utxo_key as stake,
                 enrollments.commitment,
                 enrollments.avail_height,
-                ` +
-            cur_height +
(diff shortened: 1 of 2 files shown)


MichaelKim20:
@TrustHenry 
If the block height is 20 or higher, the following PreImage values are a bit strange. Please check it out.

TrustHenry:
> @TrustHenry
> If the block height is 20 or higher, the following PreImage values are a bit strange. Please check it out.
> 
> 

It's because the pre-images have not been released.
I think that should be dealt with as a different issue.

MichaelKim20:
Agora delivers pre-image to block height 18. After that, there is no pre-image received. Don't know why.

TrustHenry:
> Agora delivers pre-image to block height 18. After that, there is no pre-image received. Don't know why.

yes!

MichaelKim20 (approved):
Endpoint `/validators` now operate normally at block heights above 20.

PR #420, opened 2021-08-30
Wallet: Convert an Unlock Bytes to a string encoded to BASE64.

MichaelKim20:
The size of the unlock bytes was changed from 64 to 65.
Therefore, I changed this to encode BASE64, one of the best ways to express bytes.


kchulj (approved):
LGTM

PR #421, opened 2021-08-31
Fix Bug: Fix field name bug in the storage of the transaction pool

MichaelKim20:
When Stoa is restarted, an exception occurs in the process of reading the previously recorded information.
Changing the field name from `tx` to `val` caused a bug in the code.
I received this from @TrustHenry. Create a correction PR for this.


TrustHenry (approved):
Thank you.

kchulj (approved):

PR #439, opened 2021-09-13
Create a new private port for agora

### LINKED ISSUE #429: Port Isolation for Trusted Endpoints
`/block_externalized`
`/block_header_updated`
`/preimage_received`
`/transaction_received`

These endpoints must be able to respond from trusted agora.
This is not public and requires that certain ports be created and protected by firewalls.

TrustHenry:
To strengthen security,
we create new port and move routing to make safe communication.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -13,7 +13,7 @@ import {
     Transaction,
     Utils,
 } from "boa-sdk-ts";
-import { cors_options } from "./cors";
+import { cors_options, cors_private_options } from "./cors";
 import { AgoraClient } from "./modules/agora/AgoraClient";
 import { IDatabaseConfig } from "./modules/common/Config";
 import { FeeManager } from "./modules/common/FeeManager";
@@ -117,11 +117,12 @@ class Stoa extends WebService {
         databaseConfig: IDatabaseConfig,
         agora_endpoint: URL,
         port: number | string,
+        private_port: number | string,
         address: string,
         genesis_timestamp: number,
         coinMarketService?: CoinMarketService
     ) {
-        super(port, address);
+        super(port, private_port, address);
 
         this.genesis_timestamp = genesis_timestamp;
         this._ledger_storage = null;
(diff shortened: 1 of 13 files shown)


src/Stoa.ts
```diff
@@ -13,7 +13,7 @@ import {
     Transaction,
     Utils,
 } from "boa-sdk-ts";
-import { cors_options } from "./cors";
+import { cors_options, cors_privte_options } from "./cors";
```
kchulj: Is this a typo? (used 3 times)

TrustHenry: Good eye!

tests/Recovery.test.ts
```diff
@@ -60,6 +60,7 @@ class TestRecoveryStoa extends TestStoa {
 describe("Test of Recovery", () => {
     const agora_addr: URL = new URL("http://localhost:2801");
     const stoa_addr: URL = new URL("http://localhost:3801");
+    const stoa_private_addr: URL = new URL("http://localhost:3802");
```
MichaelKim20:
```suggestion
    const stoa_private_addr: URL = new URL("http://localhost:4802");
```
I tried to avoid overlapping with the port of the following test code.

src/Stoa.ts
```diff
@@ -191,6 +196,21 @@ class Stoa extends WebService {
...
+                    RequesterIP: req.ip,
+                    protocol: req.protocol,
+                    httpStatusCode: res.statusCode,
+                    userAgent: req.headers["user-agent"],
+                    accessStatus: res.statusCode !== 200 ? "Denied" : "Granted",
+                    bytesTransmitted: res.socket?.bytesWritten,
+                    responseTime: time,
+                });
+            })
+        );
```
MichaelKim20, line 212: Do we really need this code?

src/modules/service/WebService.ts
```diff
@@ -47,22 +62,33 @@ export class WebService {
...
 
         this.app = express();
+        this.private_app = express();
         this.eventDispatcher = new EventDispatcher();
+
+        // Create HTTP server.
+        this.server = http.createServer(this.app);
+        this.private_server = http.createServer(this.private_app);
+
+        this._socket = new SocketIO(this.server);
```
MichaelKim20: It would be better to be created within the original start().
This is to open the socket after the route and port settings are completed rather than the creator.

MichaelKim20 (changes requested):

src/Stoa.ts
TrustHenry: it can view agora http post in the logger.

src/modules/service/WebService.ts
TrustHenry: Yes, I changed the call location.

tests/Recovery.test.ts
TrustHenry: I apply it.

TrustHenry:
@MichaelKim20 Thank you for the reivew.

MichaelKim20 (approved):
Good Work

PR #441, opened 2021-09-15
Wallet: Fix the wrong balance calculation bug when the unfreeze

MichaelKim20:


MichaelKim20:
I tested this using a wallet and confirmed the fix.

PR #444, opened 2021-09-16
Transaction socket was updated

ahmedrns:
time stamp has been added to the transaction socket method in stoa


TrustHenry (approved):

TrustHenry:
@ahmedrns Is there no unit test to test this function?

waqarRns:
> @ahmedrns Is there no unit test to test this function?

@TrustHenry meanwhile timestamp is need for the UI integration. we can proceed with unit testcases for these function later.

PR #446, opened 2021-09-17
loggers and blacklist middleware was updated

ahmedrns:
Height, status and timestamp has been added to the loggers
Reject count has been added to the blacklist middleware


src/Stoa.ts
```diff
@@ -158,8 +158,9 @@ class Stoa extends WebService {
         else {
             logger.error("LedgerStorage is not ready yet.", {
                 operation: Operation.start,
-                height: "",
-                success: false,
+                height: "0",
```
```suggestion
                height: "",
```

TrustHenry (approved):

src/Stoa.ts
ahmedrns: resolved

PR #453, opened 2021-09-25
Wallet: Fix the bug in the balance calculation

### LINKED ISSUE #452: Fix the bug on the wallet balance.
When calculating the balance of the wallet, the balance of the pending transaction was processed as locked.
However, if there is a double spent UTXO, the locked balance is deducted twice, resulting in a negative spendable balance.
We need to fill out additional test codes for each of the following cases and fix the bug.
1. When there is no pending transaction.
2. When payment transaction is pending.
3. When the freeze transaction is pending,
4. When the freeze release transition is pending,
5. When a transaction with double spent UTXO is pending,

MichaelKim20:
This was created to solve Issue #452 
Various cases of test codes have been added.
And it was confirmed that the cause of the error was the wrong SQL.
When there were multiple UTXOs in one account, the amount of the pending transaction was duplicated and locked, making Spendable Balance negative.
I solved the problem by correcting SQL accurately.


src/modules/storage/LedgerStorage.ts
```diff
@@ -2762,60 +2762,71 @@ export class LedgerStorage extends Storages {
...
-                            AND S.type != 1
-                    ), 0) AS send_other,
-                    IFNULL(
-                    (
+                            O.address = ?
+                    ) TTO
+                    INNER JOIN  (
                         SELECT
-                            SUM(O.amount)
+                            DISTINCT T.tx_hash
```
TrustHenry, line 2808: Well, Good.

TrustHenry (approved):

PR #455, opened 2021-09-28
Wallet: Improve the performance of the query on the endpoint.

MichaelKim20:
As the number of transactions increased, I found that the performance of the endpoint '/wallet/transactions/history/:address' decreased.
I create this to solve this problem.


MichaelKim20:
I checked the execution status of this in the docker's test environment.
And we confirmed that the performance has improved.
I'll merge it.

PR #456, opened 2021-09-28
Increase inquiry performance by creating index keys in the table.

MichaelKim20:
I improved the inquiry performance by generating an index key in the table.
I changed the type of address to varchar(64).


kchulj (approved):
🆗

MichaelKim20:
I applied this to the test server and tried it on the wallet.
The inquiry performance has improved a lot.

PR #467, opened 2021-10-12
fix: Transaction type

ahmedrns:
fix: Holder balance history API
fix: Average fee char API
fix: tx_fee, tx_size and full_count was added to wallet transaction history


src/Stoa.ts
```diff
@@ -57,7 +57,7 @@ import {
     IProposalList,
     IValidatorReward
 } from "./Types";
-
+import _ from 'lodash';
```
TrustHenry, line 60: https://github.com/bosagora/stoa/blob/6f5026a59788c5f9206ebf5b8e69890b2e8cdf2c/src/Types.ts#L558

```
static tx_types: string[] = ["Payment", "Freeze", "Coinbase"];
static display_tx_type: string[] = ["Inbound", "Outbound", "Freeze", "Payload"];
```

You can change it like this.
If so, you don't have to convert it into capital letters.

ahmedrns: I have tired this one, but it causes errors with wallet transaction history test, I think it will not work with filter types

TrustHenry (approved):

src/Stoa.ts
TrustHenry: I checked the code that affects the web wallet.

PR #470, opened 2021-10-13
Add endpoint of wallet /wallet/transaction/detail/:hash

MichaelKim20:
It is similar to /wallet/transaction/overview/:hash and was created to rename.
After this is completed, /wallet/transaction/overview/:hash will be removed.


TrustHenry (approved):
LGTM

PR #472, opened 2021-10-14
fix: CoinMarket Service was fixed

ahmedrns:
fix: boa-stat API was updated

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -836,9 +836,9 @@ class Stoa extends WebService {
         filter_type =
             req.query.type !== undefined
                 ? req.query.type
-                      .toString()
-                      .split(",")
-                      .map((m) => ConvertTypes.toDisplayTxType(m))
+                    .toString()
+                    .split(",")
+                    .map((m) => ConvertTypes.toDisplayTxType(m))
                 : [0, 1, 2, 3];
 
         if (filter_type.find((m) => m < 0) !== undefined) {
@@ -1303,7 +1303,8 @@ class Stoa extends WebService {
                         height: data[0].height,
                         transactions: data[0].transactions,
                         validators: data[0].validators,
-                        frozen_coin: 5283595, // FIX ME static data because of unavailability of real data
+                        frozen_coin: data[0].total_frozen,
+                        total_reward: data[0].total_reward,
                         circulating_supply: 5283535,
                         active_validators: 155055,
                     };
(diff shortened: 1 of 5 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -3501,11 +3501,14 @@ export class LedgerStorage extends Storages {
...
-            FROM
-                blocks;`;
+        const sql = `
+                SELECT max(height) as height,
+	                (SELECT SUM(total_reward) as total_reward from accounts) as total_reward,
+                    (SELECT SUM(total_frozen) as total_frozen from accounts) as total_frozen,
+                    (SELECT count(*) from transactions) as transactions,
+                    (SELECT count(DISTINCT address) from validators) as validators
+                    FROM
+                    blocks;`;
```
TrustHenry: It seems like a query that lot of costs.
Can you reduce the DB load?
Is aggregate data not available?

src/Stoa.ts
```diff
@@ -2081,9 +2082,8 @@ class Stoa extends WebService {
                     if (changes)
                         logger.info(
                             `Saved a pre-image utxo : ${pre_image.utxo.toString().substr(0, 18)}, ` +
-                                `hash : ${pre_image.hash.toString().substr(0, 18)}, pre-image height : ${
-                                    pre_image.height
-                                }`,
+                            `hash : ${pre_image.hash.toString().substr(0, 18)}, pre-image height : ${pre_image.height
```
TrustHenry: Is there a reason why you use `substr`?
Why don't we loged the entire hash?

ahmedrns: This was used before, in fact entire hash can be logged

ahmedrns: Entire hash is now being logged

src/modules/storage/LedgerStorage.ts
```diff
@@ -3501,11 +3501,16 @@ export class LedgerStorage extends Storages {
...
-             (SELECT count(*) from transactions) as transactions,
-             (SELECT count(*) from validators) as validators
-            FROM
-                blocks;`;
+        const sql = `
+                SELECT MAX(B.height) as height,
+                    SUM(B.tx_count) as transactions,
+	                SUM(BS.total_reward) as total_reward,
+                    (SELECT SUM(total_frozen) as total_frozen from accounts) as total_frozen,
+                    (SELECT COUNT(DISTINCT address) from validators) as validators
```
TrustHenry, line 3509: I don't think this value is right.
You must count the number of valid validators for the corresponding block height. Right?

ahmedrns: Active validators are yet to be added, these are the total number of the validators of the system, active validators will be fetched on every block height

TrustHenry: This part needs to be corrected later.

TrustHenry (approved):

PR #474, opened 2021-10-18
When data does not exist, it delivers an empty array instead of an error code

MichaelKim20:
It delivers code 204 when data does not exist.
I modified this to transfer the arrangement.


TrustHenry (approved):
LGTM

PR #475, opened 2021-10-20
Add the ability to create blocks, validator, proposal, vote, counting in the test code.

MichaelKim20:
The validators can be added and removed from the test code.
Proposals, voting, and counting can be made in the test code, so tests can be conducted for various types.


TrustHenry (approved):
Awesome!

PR #481, opened 2021-10-26
BlockManager: Change the interval that becomes validator after enrollment in the test code

MichaelKim20:
I changed the code so that the validator can be from the block with registration information.
I'm going to check it out in Agora and record the results.

### DIFF
--- a/tests/Congress.test.ts
+++ b/tests/Congress.test.ts
@@ -148,7 +148,7 @@ describe("Test for the addition of validators", () => {
         // Create a new block.
         const new_block = block_manager.saveBlock([tx], enrollments);
 
-        const expected_validator = block_manager.validators.get(block_manager.height - 1).map((key) => key.address);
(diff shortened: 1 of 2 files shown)


MichaelKim20:
I modified the blockhead information to be output to the test code that already exists.

**Test Code**
https://gist.github.com/MichaelKim20/7ee87c32f1e961256d2325e67823dc6a#file-manyvalidators-d

**Print**
https://gist.github.com/MichaelKim20/7ee87c32f1e961256d2325e67823dc6a#file-result-txt

**Result**
Enrollment was added to height 20. 
However, in the block header, it was added as a verifier from block 21.

MichaelKim20:
Accordingly, the new validator may be a validator from the next block including the enrollment.

MichaelKim20:
**Condition:**
At block height 1, five enrollment were added.

**Before this PR:**


**After this PR:**


Therefore, this PR should be merged to operate in the same way as Agora.

TrustHenry (approved):

PR #482, opened 2021-11-01
Fix the order of validators

TrustHenry:
A meaningless sorting order was added.
Now it's only sorted with utxo_key.


MichaelKim20 (approved):
LGTM

PR #484, opened 2021-11-03
TransactionPool: Change according to Agora's change

MichaelKim20:
I created this PR for solving issue https://github.com/bosagora/stoa/issues/464
Even if this is not solved, it is not a big problem.
However, we prepare in advance to minimize the impact of TxInput.unlock_age.


TrustHenry (approved):
LGTM

PR #485, opened 2021-11-03
WalletWatcher: Implementation the WalletWatch

MichaelKim20:


src/modules/wallet/WalletWatcherIO.ts
```diff
@@ -0,0 +1,94 @@
...
+                const found = values.findIndex((m) => m === client_socket);
+                if (found >= 0) values.splice(found, 1);
+            }
+        });
+
+        client_socket.on("ping", (data: any) => {
+            this.ping.set(client_socket.id, new Date().getTime());
+        });
+
+        client_socket.on("disconnect", () => {
```
TrustHenry, line 103: If the client shuts down abnormally without calling disconnect, garbage sockets may accumulate.

MichaelKim20: I added the following functions.
The client sends a ping to the server once every 30 seconds.
The server calls the function checkConnection() every minute.
This function disconnects a client whose ping has been latest received for more than one minute.
And remove it from the subscription table.

TrustHenry (approved):
LGTM

PR #486, opened 2021-11-04
Pending transaction was added to /latest-transactions API

waqarRns:


src/modules/storage/LedgerStorage.ts
```diff
@@ -3442,15 +3442,44 @@ export class LedgerStorage extends Storages {
...
+                    SELECT
+                        (SELECT IFNULL(MAX(height), 0) AS height FROM blocks) AS block_height,
+	                    T.tx_hash,
+                        T.time as time_stamp,
+	                    T.tx_fee,
+                        T.tx_size,
+                        T.type,
+	                    Sum(IFNULL(O.amount,0)) as amount
+	                FROM
+	                    tx_outputs O
```
TrustHenry, line 3465:
```suggestion
	                    tx_output_pool O
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -3442,15 +3442,44 @@ export class LedgerStorage extends Storages {
...
+                    amount,
+                    count(*) OVER() AS full_count
+                FROM(
+                    SELECT
+                        (SELECT IFNULL(MAX(height), 0) AS height FROM blocks) AS block_height,
+	                    T.tx_hash,
+                        T.time as time_stamp,
+	                    T.tx_fee,
+                        T.tx_size,
+                        T.type,
```
TrustHenry, line 3461: '0' as status,
// Pending

src/modules/storage/LedgerStorage.ts
```diff
@@ -3442,15 +3442,44 @@ export class LedgerStorage extends Storages {
...
+		                INNER JOIN transaction_pool T ON (T.tx_hash = I.tx_hash)
+                    UNION
+                    SELECT
+                        T.block_height,
+                        T.tx_hash,
+                        B.time_stamp,
+                        T.tx_fee,
+                        T.tx_size,
+                        T.type,
+                        Sum(IFNULL(O.amount,0)) as amount
```
TrustHenry: '1' as status,
// Confirmed

TrustHenry (changes requested):

src/modules/storage/LedgerStorage.ts
waqarRns: we can not use tx_output_pool. actully we need 
tx_outputs S
                INNER JOIN tx_input_pool I ON (I.utxo = S.utxo_key)

src/modules/storage/LedgerStorage.ts
```diff
@@ -3442,15 +3442,48 @@ export class LedgerStorage extends Storages {
...
+                    UNION
+                    SELECT
+                        T.block_height,
+                        T.tx_hash,
+                        B.time_stamp,
+                        T.tx_fee,
+                        T.tx_size,
+                        T.type,
+                        Sum(IFNULL(O.amount,0)) as amount,
+                        'Confirmed' as status
```
TrustHenry, line 3478: How about coding with multilingual processing on the screen?

waqarRns: we cannot apply multilingual processing on dynamic data.
i think we should only apply it to static data.

TrustHenry (approved):

PR #489, opened 2021-11-08
Increase the delay time of the test code.

MichaelKim20:
Create this to solve the problem that tests are not performed normally due to a lack of delay time in CI.


TrustHenry (approved):
Try

PR #493, opened 2021-11-09
Add the sorting order of /latest-blocks

TrustHenry:
Sort in the order of tx_hash asc at the same timestamp.


MichaelKim20 (approved):
LGTM

PR #494, opened 2021-11-09
Improve WalletWatcher

MichaelKim20:
1. Duplicate events have been removed for one address.
2. Events that occur when creating blocks are removed to prevent load concentration.
3. Additional information is added to the event.
    - Transaction hash: Hash of a transaction that caused a change in balance
    - Type
         *  "pending": In case of a transaction that is not yet included in the block.
         *  "confirm": When a block is created, when it is a transaction included in the block.

### DIFF
--- a/src/Stoa.ts
+++ b/src/Stoa.ts
@@ -2145,7 +2145,6 @@ class Stoa extends WebService {
                         });
                         await this.emitBlock(block);
                         await this.emitBoaStats();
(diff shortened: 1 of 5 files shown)


TrustHenry (approved):
LGTM

TrustHenry:
Conflicting files
tests/Congress.test.ts

PR #495, opened 2021-11-09
Slashed validators were applied

### LINKED ISSUE #487: preimages table default key constraint violation
when re-enrollment is done at height 20. then preimages at height 21  violation the primary key constraint

.

TrustHenry:
If there is no pre-image in the block header, it is slashed.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -41,13 +41,24 @@ import {
     UnspentTxOutput,
     Utils,
     UTXOManager,
-    VarInt
+    VarInt,
 } from "boa-sdk-ts";
 import bigDecimal from "js-big-decimal";
 import moment from "moment";
 import * as mysql from "mysql2";
 import { SmartBuffer } from "smart-buffer";
-import { IAccountInformation, IBallot, IMarketCap, IMetaData, IPreimage, IProposal, IProposalAttachment, IValidator, IValidatorByBlock, IVotingResult } from "../../Types";
+import {
+    IAccountInformation,
+    IBallot,
+    IMarketCap,
+    IMetaData,
+    IPreimage,
+    IProposal,
+    IProposalAttachment,
+    IValidator,
+    IValidatorByBlock,
+    IVotingResult,
+} from "../../Types";
 import { IDatabaseConfig } from "../common/Config";
 import { DataCollectionStatus, ProposalResult, ProposalStatus, SignStatus, ValidatorStatus } from "../common/enum";
 import { FeeManager } from "../common/FeeManager";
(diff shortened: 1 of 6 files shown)


src/modules/storage/LedgerStorage.ts
```diff
@@ -907,20 +907,54 @@ export class LedgerStorage extends Storages {
...
         }
 
+        function updateValidatorSlashing(storage: LedgerStorage, height: Height, utxo_key: Hash, enrolled_at: Height) {
+            return new Promise<void>((resolve, reject) => {
+                storage
+                    .query(
+                        `UPDATE validators
+                            SET slashed = ? ,
+                            slashed_height = ?
+                            WHERE utxo_key = ? AND enrolled_at = ?`,
```
MichaelKim20, line 941: It changed from `address` to `utxo_key`.
Good

src/modules/storage/LedgerStorage.ts
```diff
@@ -907,20 +907,54 @@ export class LedgerStorage extends Storages {
...
         return new Promise<void>((resolve, reject) => {
             (async () => {
                 const validators: any[] = await this.getValidatorsAPI(block.header.height, null, conn);
                 for (let idx = 0; idx < block.header.preimages.length; idx++) {
+                    if (idx >= validators.length) continue;
+                    if (validators[idx] === undefined) continue;
                     try {
-                        if (undefined !== validators[idx] && block.header.preimages[idx] !== new Hash(Buffer.alloc(Hash.Width))) {
-                            await save_preimage(
+                        if (Buffer.compare(block.header.preimages[idx].data, Buffer.alloc(Hash.Width)) === 0) {
```
```suggestion
                        if (Hash.isNull(block.header.preimages[idx])) {
```
Please raise the version of SDK to 0.2.6, and I hope you can modify the code like this.

src/modules/storage/LedgerStorage.ts
```diff
@@ -2888,75 +2898,25 @@ export class LedgerStorage extends Storages {
...
-                slashedValidators = cycleValidators.filter(v => !block_preimages.some((v1) => v.address === v1.address));
-
-                for (let i = 0; i < slashedValidators.length; i++) {
-                    await updateValidatorSlashing(this, block.header.height, slashedValidators[i].address, slashedValidators[i].enrolled_at);
-                }
-
-                slashedValidators.forEach((e) => {
-                    cycleValidators.splice(cycleValidators.indexOf(e), 1);
-                });
+                const cycleValidators: any[] = await this.getValidatorsAPI(block.header.height, null);
```
```suggestion
                const cycleValidators: any[] = await this.getValidatorsAPI(block.header.height, null, conn);
```

MichaelKim20 (changes requested):
I've reviewed it so far.
Let me check more.

src/modules/storage/LedgerStorage.ts
TrustHenry: I applied the suggestion.

src/modules/storage/LedgerStorage.ts
TrustHenry: I applied the suggestion.

MichaelKim20: The suggestion above is wrong.
I think it'd be good to do it like this.
```suggestion
                        if (block.header.preimages[idx].isNull()) {
```

TrustHenry: I applied the suggestion.

MichaelKim20 (approved):
LGTM

PR #502, opened 2021-11-18
Remove `time_offset` in the block

MichaelKim20:
In the production environment, it should be treated as the received time.


TrustHenry (approved):
LGTM

PR #505, opened 2021-11-22
timestamp issue in Charts has been fixed

waqarRns:


src/Types.ts
```diff
@@ -1814,6 +1819,11 @@ export interface IAccountChart {
      */
     granularity: string;
 
+    /**
+     * Granularity of Record
```
TrustHenry:
```suggestion
     * Granularity timestamp of Record
```

PR #512, opened 2021-12-02
circulating supply was updated

ahmedrns:


src/modules/storage/LedgerStorage.ts
```diff
@@ -1527,12 +1538,7 @@ export class LedgerStorage extends Storages {
                                             WHERE
                                                 block_height =?;`;
 
-                const circulating_supply_sql = `SELECT
-                                                circulating_supply
-                                            FROM
-                                            blocks_stats
-                                            WHERE
-                                                block_height = ?;`;
+                const get_utxo = `SELECT SUM(amount) as total_utxo FROM utxos`;
```
```suggestion
                const get_sum_utxos_sql = `SELECT SUM(amount) as total_utxo FROM utxos`;
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -1543,24 +1549,32 @@ export class LedgerStorage extends Storages {
...
                             default: {
                                 total_received = JSBI.BigInt(row[0].total_amount);
-                                if (Number(block.header.height) === 0) {
-                                    circulating_supply = total_received;
-                                }
                                 break;
                             }
                         }
-                        return this.query(circulating_supply_sql, [Number(block.header.height) - 1], conn);
+                        return this.query(get_utxo, [], conn);
```
```suggestion
                        return this.query(get_sum_utxos_sql, [], conn);
```

src/modules/storage/LedgerStorage.ts
```diff
@@ -1543,24 +1549,32 @@ export class LedgerStorage extends Storages {
...
+                            total_utxo = JSBI.BigInt(row[0].total_utxo);
+                            let excludedAddresses: string[]
+                            if (this.excluded_addresses !== undefined) {
+                                excludedAddresses = this.excluded_addresses
+                            } else {
+                                excludedAddresses = []
+                            }
+                            await Promise.all(excludedAddresses.map(async (elem: string) => {
+                                const balance: any = await this.getWalletBalance(elem.toString());
+                                total_common_budget_balance = JSBI.add(total_common_budget_balance, JSBI.BigInt(balance[0].balance))
```
TrustHenry: `this.getWalletBalance` does not reflect data on new blocks because it is not the same transaction DB connection.

`balance[0].balance`  <- Index reference errors may occur.

tests/Boascan.test.ts
```diff
@@ -721,7 +721,7 @@ describe("Test of Stoa API Server", function () {
             validators: 6,
             frozen_coin: '120000000000000',
             total_reward: '0',
-            circulating_supply: "4999999998096000",
+            circulating_supply: 4999999998096000,
```
TrustHenry, line 724: Why did you convert it into a number without returning it to a string?

src/modules/storage/LedgerStorage.ts
TrustHenry: How about referring to it at the accounts table?

TrustHenry: If the balance of the accounts table of the corresponding block height has been calculated, it may be referred to.

TrustHenry: Or if you use 'this.getWalletBalance', you need to add a connection.

src/modules/storage/LedgerStorage.ts
```diff
@@ -4730,7 +4744,7 @@ export class LedgerStorage extends Storages {
      * and if an error occurs the .catch is called with an error.
      */
     public getBOAHolders(limit: number, page: number): Promise<any> {
-        const circulating_sql = `select sum(amount) as circulating_supply from utxos`;
+        const circulating_sql = `SELECT circulating_supply FROM blocks_stats WHERE block_height = (SELECT MAX(block_height) FROM blocks_stats)`;
```
TrustHenry, line 4749: Can't you combine it with the query below?

src/modules/storage/LedgerStorage.ts
```diff
@@ -5105,7 +5119,7 @@ export class LedgerStorage extends Storages {
      * and if an error occurs the .catch is called with an error.
      */
     public getBOAHolder(address: string): Promise<any> {
-        const circulating_sql = `select sum(amount) as circulating_supply from utxos;`;
+        const circulating_sql = `SELECT circulating_supply FROM blocks_stats WHERE block_height = (SELECT MAX(block_height) FROM blocks_stats)`;
```
TrustHenry, line 5124: Can't you combine it with the query below?

src/modules/storage/LedgerStorage.ts
ahmedrns: I don't think there's any common column, If i combine in the query blow it will be a nested query

src/modules/storage/LedgerStorage.ts
ahmedrns: I have added a connection,

tests/Boascan.test.ts
ahmedrns: In IBOAStats interface, the circulating supply is number type

src/modules/storage/LedgerStorage.ts
TrustHenry: balance[0].balance <- Index reference errors may occur.

ahmedrns: check has been implemented on balance[0]

TrustHenry (approved):
LGTM

PR #515, opened 2021-12-08
Change the method of calculating the block timestamp

TrustHenry:
Agora changed the method of calculating the block timestamp.

### DIFF
--- a/src/modules/storage/LedgerStorage.ts
+++ b/src/modules/storage/LedgerStorage.ts
@@ -585,13 +585,7 @@ export class LedgerStorage extends Storages {
     public putBlocks(block: Block): Promise<void> {
         const genesis_timestamp: number = this.genesis_timestamp;
         const now_timestamp = Math.floor(new Date().getTime() / 1000);
-
-        // In the production environment, it should be treated as the received time.
-        if (process.env.NODE_ENV === "production") {
-            block.header.time_offset = now_timestamp - genesis_timestamp;
-        } else {
-            block.header.time_offset = JSBI.toNumber(block.header.height.value) * this.block_interval;
-        }
+        block.header.time_offset = JSBI.toNumber(block.header.height.value) * this.block_interval;
 
         function saveBlock(
             storage: LedgerStorage,


MichaelKim20:
If you expect a timestamp to be included in the block next time, it seems that there are too many changes.
If the timestamp is not included in the future, I think it's better to decide whether the block is newly created or restored.

src/modules/storage/LedgerStorage.ts
```diff
@@ -583,20 +590,13 @@ export class LedgerStorage extends Storages {
...
         const now_timestamp = Math.floor(new Date().getTime() / 1000);
-
-        // In the production environment, it should be treated as the received time.
-        if (process.env.NODE_ENV === "production") {
-            block.header.time_offset = now_timestamp - genesis_timestamp;
-        } else {
-            block.header.time_offset = JSBI.toNumber(block.header.height.value) * this.block_interval;
-        }
-
+        block.header.time_offset = now_timestamp - this.genesis_timestamp;
```
hewison-chris: What will this time_offset be used for?

hewison-chris: I think it should also be set to the calculated offset based on the block height.

MichaelKim20:
> If you expect a timestamp to be included in the block next time, it seems that there are too many changes. If the timestamp is not included in the future, I think it's better to decide whether the block is newly created or restored.

Of course, this is what Henry and I decided to do on RLT.

MichaelKim20:
Agora will always make the blocks exactly at time intervals, but assuming that exceptional situations can arise, I suggest the following way.

Case 1) When a new block is received:

	block.header.time_offset = Math.floor(new Date().getTime() / 1000) - genesis_timestamp;

Case 2) When a recovery block is received:

        Height of recent block: LH
        TimeStamp of recent block: LT

        Height of recovery block: RH
        TimeStamp of recovery block: RT

        RT = (RH - LH) * Block_Interval + LT
        block.header.time_offset = RT - genesis_timestamp;


경우 1) 신규 블럭이 수신되었을 때 :

	block.header.time_offset = Math.floor(new Date().getTime() / 1000) - genesis_timestamp;
	
경우 2) 복구 블럭이 수신되었을 때 :

	최근 블록의 높이 : LH
	최근 블록의 타임스탬프 : LT
	
	복구 블록의 높이 : RH
	복구 블록의 타임스탬프 : RT

	RT = (RH - LH) * Block_Interval + LT

	block.header.time_offset = RT - genesis_timestamp;

TrustHenry:
@hewison-chris I'd like to hear @hewison-chris' opinion on Michael's comment.

hewison-chris:
I still think that we should use a consistent timestamp for blocks that is only calculated from the block height. If we start creating timestamps based on the clock time on different servers which might receive the block at different times it will lead to confusion. When you think about it the actual time a block is externalized can be different on every node. The transactions that are included were posted at various times during the block interval so each `tx` has in theory a different timestamp, but we have a general timestamp which is based on the intended block interval.

MichaelKim20:
> I still think that we should use a consistent timestamp for blocks that is only calculated from the block height. If we start creating timestamps based on the clock time on different servers which might receive the block at different times it will lead to confusion. When you think about it the actual time a block is externalized can be different on every node. The transactions that are included were posted at various times during the block interval so each `tx` has in theory a different timestamp, but we have a general timestamp which is based on the intended block interval.

Your suggestion makes sense. First, let's revise the code according to your opinion. And let's keep checking in the future.

TrustHenry:
I applied @hewison-chris @MichaelKim20 as a suggestion.

hewison-chris (approved):
LGTM

MichaelKim20 (approved):
LGTM
You've done a lot of hidden work.

TrustHenry:
Thank you for the review.

PR #524, opened 2021-12-21
Freezing Fee: Apply the freezing fee to the class LedgerStorage

MichaelKim20:
Freezing TXs now have to include SlashAmountPenatly amount of fee for each Freeze output they have
It's a related PR in Agora.
bosagora/agora#2597

I added a data field where the freeze fee will be stored.
I modified the code to calculate the freeze fee in the transaction and store it in the database.
I did the same thing not only in the transactions included in the block but also in the fandom transaction.
The remaining issue is to reflect the freezing deposit included in the frozen UTXO in the account balance.
I'm going to solve this in a new issue.


Kang-Simon (approved):
LGTM

TrustHenry (approved):
Nice!

PR #534, opened 2022-02-16
Fix test code of several files

MichaelKim20:
I fixed the test code of several files.

The test data contained many transactions.
So I used a larger delay time to give enough time for the data to be processed.


TrustHenry (approved):
Nice!

PR #544, opened 2022-04-18
Implement `/spender/:utxo` endpoint

omerfirmak:


hewison-chris (approved):
LGTM, though we should probably add a unit test for it.

omerfirmak:
> LGTM, though we should probably add a unit test for it.

Done

